PR #20, opened 2017-03-15
Add "count" attribute to reverse-related

tritoch:
Cards that create multiple of a single type of token should be designated as such for future functionality.

ex: `Lingering Souls` creates 2 spirit tokens.  It now is stored as: `<reverse-related count="2">Lingering Souls</reverse-related>`

### DIFF
--- a/tokens.xml
+++ b/tokens.xml
@@ -376,7 +376,7 @@
             <text>Flying</text>
             <token>1</token>
             <reverse-related>Battle Screech</reverse-related>
-            <reverse-related>Beck // Call</reverse-related>
+            <reverse-related count="4">Beck // Call</reverse-related>
             <reverse-related>Emeria Angel</reverse-related>
             <reverse-related>Jötun Owl Keeper</reverse-related>
             <reverse-related>Migratory Route</reverse-related>
@@ -4367,7 +4367,7 @@ Sacrifice this creature: This creature deals 1 damage to target creature or play
             <reverse-related>Empty the Pits</reverse-related>
             <reverse-related>Endless Ranks of the Dead</reverse-related>
             <reverse-related>Feast or Famine</reverse-related>
-            <reverse-related>From Under the Floorboards</reverse-related>
+            <reverse-related count="3">From Under the Floorboards</reverse-related>
             <reverse-related>Ghoulcaller Gisa</reverse-related>
             <reverse-related>Ghoulcaller's Accomplice</reverse-related>
             <reverse-related count="2">Gisa's Bidding</reverse-related>
--- a/tokens.xml
+++ b/tokens.xml
@@ -33,7 +33,7 @@
             <set picURL="http://magiccards.info/extras/token/conflux/angel.jpg">CON</set>
             <set picURL="http://magiccards.info/extras/token/player-rewards-2004/angel.jpg">SCG</set>
             <color>W</color>
-            <manacost></manacost>
+            <manacost />
             <type>Token Creature — Angel</type>
             <pt>4/4</pt>
             <tablerow>2</tablerow>
@@ -56,7 +56,7 @@
             <name>Angel </name>
             <set picURL="http://cdn.staticneo.com/w/mtg/b/bb/Angel3.jpg">APC</set>
             <color>B</color>
-            <manacost></manacost>
+            <manacost />
             <type>Token Creature — Angel</type>
             <pt>3/3</pt>
             <tablerow>2</tablerow>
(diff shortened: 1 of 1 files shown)


ZeldaZach:
This is an interesting change, I like it 👍 
@Psithief What do you think?

Psithief:
Seems bloody hard, and doesn't cover cards that have two numbers about creating tokens, such as the following:

```
Gather the Townsfolk

Create two 1/1 white Human creature tokens.
Fateful hour — If you have 5 or less life, create five of those tokens instead.
```

```
Increasing Devotion

Create five 1/1 white Human creature tokens. If Increasing Devotion was cast from a graveyard, create ten of those tokens instead.
Flashback {7}{W}{W} (You may cast this card from your graveyard for its flashback cost. Then exile it.)
```

```
Throne of Empires

{1}, {T}: Create a 1/1 white Soldier creature token. Create five of those tokens instead if you control artifacts named Crown of Empires and Scepter of Empires.
```

There's no fixed pattern here. It could be on a separate line, or the same line. It could be at the beginning of the sentence or the end.

It doesn't cover cards that create a number of different tokens as a set, such as these:

```
Bestial Menace

Create a 1/1 green Snake creature token, a 2/2 green Wolf creature token, and a 3/3 green Elephant creature token.
```

```
Trostani's Summoner

When Trostani's Summoner enters the battlefield, create a 2/2 white Knight creature token with vigilance, a 3/3 green Centaur creature token, and a 4/4 green Rhino creature token with trample.
```

Psithief:
I think the biggest problem here is the implication that Cockatrice would take away the ability to create **one** of any related token and only allow creating multiples if that's what the card says.

That's really a UI discussion that needs to happen in Cockatrice/Cockatrice first.

tritoch:
>doesn't cover cards that have two numbers about creating tokens, such as the following

It always takes the smallest number.

>It doesn't cover cards that create a number of different tokens as a set, such as these:

Sure it does.  Bestial Menace, Trostani's Summoner makes 1 of each of those tokens.  1 is the assumed number if count is absent.  Why would you put a count in the Wolf or Knight reverse-related for those cards?

>That's really a UI discussion that needs to happen in Cockatrice/Cockatrice first.

https://github.com/Cockatrice/Cockatrice/issues/2262

tooomm:
>I think the biggest problem here is the implication that Cockatrice could take away the ability to create one of any related token and only allow creating multiples if that's what the card says.

Can you explain why that might be an issue? Gladly on the trice issue.

Psithief:
A "count" attribute doesn't address how to deal with cards that create a variable amount of tokens.
From Under the Floorboards might read as `<reverse-related count="x=3">From Under the Floorboards</reverse-related>` and have a create X tokens dialog only that defaults to 3.

We may also need more attributes, such as an "exclude" attribute to prevent "All Tokens" doing its operation blindly. 

"All Tokens" also won't save any of the tokens created to "Create another token". Do we want a "default" attribute for that so it always picks something? Or should "Create another token" be able to create more than one token (of different types) at a time?

tooomm:
>From Under the Floorboards might read as <reverse-related count="x=3">From Under the Floorboards</reverse-related> and have a create X tokens dialog only that defaults to 3.

That's a very sweet solution for that card!

As @tritoch stated above only catching the lowest number is a working solution for most cards I guess. But if there is an `X` involved, it should be `"x=lowest number"` or just x which defaults to 1.

Psithief:
Maybe `count="~3"` for 'x' with a default value would be simpler syntax.

tooomm:
Hmm, this looks like really really confusing to me.
We should have a `x` somewhere in there for sure!

Since x can be any value there is no need for a variable default?
It should always default to the lowest possible (which is 1, because for 0 you don't want to hit the option to create something at all - you might just say it in chat)

Psithief:
Sorry, I was still talking about `From Under the Floorboards`. :+1: :grin:

tooomm:
I think this card should have obviously two reverse related tokens?
Like a planeswalker etc. it has two options to create tokens, just happens to be the same type of tokens
 - x Zombies
- 3 Zombies

Psithief:
That's a good idea, but will it create havoc with the current "create all tokens" shortcut? (e.g. `gather the townsfolk`)
I think that means this solution needs another attribute to exclude that shortcut from counting any particular entry.

tooomm:
Good argument... hmm.
I don't like the `create all token` feature too much anyway since it messes around with most cards like all planeswalkers and there are just a few cards which actually make good use out of it.
Maybe 5% of all token creators? Most likely less..

If there is an x involved, it should not be picked up by `create all`, but in your example it's different.
Maybe prevent that feature for token creators which have several options to create the same token?

So much effort needed to make `create all` work properly for very few cards witout messing around with the majority... (or prevent useful features for everybody, there are way more x-creators than create 3 different token at once!)

tritoch:
See updated #38

PR #24, opened 2017-04-20
fixed typo

tooomm:
fixing https://github.com/Cockatrice/Cockatrice/issues/2612


Psithief:
Only a single typo? That's not bad.

PR #33, opened 2017-05-03
Fixed Dissension and Conflux Elemental image

card-collector:
The image set for Dissension was the image from Conflux.  I found the Dissension image at the Neoseeker wiki (where some other Dissension images were linked from).


Psithief:
Actually, if there is a source from cdn.staticneo and a genuine source, it's better to delete the old one.
Dissension did not have printed tokens. They are fan-made.

PR #34, opened 2017-05-03
Added second Shadowmoor Elemental token

card-collector:


Psithief:
This token is **intentionally** omitted due to Wizard's policy of not printing "haste" on tokens as of `Avacyn Restored`.

PR #35, opened 2017-05-03
Updated Elves vs. Goblins tokens

card-collector:
The images for the Anthologies version were the images for the original Duel Deck.


card-collector:
After looking around there doesn't seem to be a consistent set code for the 2 versions, the original setup might actually be correct.

PR #37, opened 2017-05-27
Add Morph & Manifest Reverse-Related

tritoch:
Fixes https://github.com/Cockatrice/Cockatrice/issues/2733


Psithief:
Did a large amount of people ask for this? It seems like unnecessary clutter, given how broad turning a card face down is.

Psithief:
After further thought, marking as `won't fix`. Morph and manifest are not tokens. These entries are just markers. I would not want to have a "create all tokens" shortcut making unattached morph reminders.

PR #38, opened 2017-06-15
Add "count" attribute for multiple tokens

tritoch:
Adds count attribute for cards that create multiple tokens. Includes `X` variable creators.


ZeldaZach:
How'd you generate this exactly?

Psithief:
I _sort of_ just pushed my own version. :1st_place_medal:

tooomm:
Oh wow, so bad that we work on different fronts here all the time and do work twice...

Can you explain your solution and what you did different and why?
I realized differences for cards like "From Under the Floorboards" (fix option + x option) and "Increasing Devotion" (two fix options). As well as more additions and less deletions, haha.

I also saw that you also add the other things like `attach` and `exclude` within the same commit (https://github.com/Cockatrice/Magic-Token/commit/c2fdecc53c954e3ae9b82b7880f82c99ef380540).

Psithief:
1.  I wanted to have as few unnecessary changes as possible.
2.  I wanted to see if my approach worked, and if I missed any, or you missed any and why.

So I'm using your pull request to check my own approach.

It looks like you only tagged tokens in the form "create X", but missed the "for each __ create" and "create ___ token ___ for each" forms.

And that's pretty annoying to catch programmatically, so I did it manually.

PR #45, opened 2017-09-26
wizards to https

tooomm:
wizards finally supports https!

### DIFF
--- a/tokens.xml
+++ b/tokens.xml
@@ -35,9 +35,9 @@
         </card>
         <card>
             <name>Angel</name>
-            <set picURL="http://media.wizards.com/2017/b9jm8nwv_mm3/en_gvwFXo4kL2.png">MM3</set>
-            <set picURL="http://media.wizards.com/2016/aksdjciawolkcc0_soi/en_OnvtmxOuug.png">SOI</set>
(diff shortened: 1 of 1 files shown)


Psithief:
Supports? or is it required? They are very different terms.

tooomm:
>Supports?

Yes.

>or is it required?

No.

>They are very different terms.

Yes.

>The difference is rather important.

Yes.

Nothing was broken before this change, and nothing will brake after it.
Wizards only managed to allow secure connections in between...

Psithief:
The connection overhead for the security provided by https is not warranted, since these are only image files. If files can still be downloaded by http, we shouldn't change this.

tooomm:
So you want to wait until unsecured http is fully deprecated instead building future proof now?

What's about the 472 https links to mci then? ;)

Psithief:
Once http becomes fully deprecated on the server, what will happen is that the http request will become redirected to a https request. This already happens for magiccards.info, so using the https there is faster.

ZeldaZach:
I see no reason to avoid merging this PR. Always better to secure a link.

Psithief:
No Zach, that is a completely pointless waste of extra electrical power for no benefit to anyone. There is no meaningful communication to secure. These are _public, non-executable resources_.

PR #51, opened 2017-12-14
Remove photobucket URLs.

### LINKED ISSUE #50: another photobucket link broken
Just like https://github.com/Cockatrice/Magic-Token/issues/49, same problem with:
https://github.com/Cockatrice/Magic-Token/blob/master/tokens.xml#L1105

I double checked all photobucket links because of https://github.com/Cockatrice/Cockatrice/issues/2955.

Psithief:
Fix #50

### DIFF
--- a/tokens.xml
+++ b/tokens.xml
@@ -923,7 +923,7 @@ Cloud Sprite can block only creatures with flying.</text>
         </card>
         <card>
             <name>Deserter</name>
-            <set picURL="https://i.imgur.com/cHNbJQM.jpg    ">ME2</set>
+            <set picURL="https://i.imgur.com/cHNbJQM.jpg">ME2</set>
             <color>W</color>
             <cmc>0</cmc>
             <type>Token Creature — Deserter</type>
--- a/tokens.xml
+++ b/tokens.xml
@@ -923,7 +923,7 @@ Cloud Sprite can block only creatures with flying.</text>
         </card>
         <card>
             <name>Deserter</name>
-            <set picURL="http://i1013.photobucket.com/albums/af260/lovesoldier99/DESERTERTOKEN.jpg">ME2</set>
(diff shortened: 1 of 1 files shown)


tokens.xml
```diff
@@ -923,7 +923,7 @@ Cloud Sprite can block only creatures with flying.</text>
         </card>
         <card>
             <name>Deserter</name>
-            <set picURL="http://i1013.photobucket.com/albums/af260/lovesoldier99/DESERTERTOKEN.jpg">ME2</set>
+            <set picURL="https://i.imgur.com/cHNbJQM.jpg    ">ME2</set>
```
tooomm: Spaces

tokens.xml
```diff
@@ -2049,7 +2049,7 @@ Whenever Glyph Keeper becomes the target of a spell or ability for the first tim
         </card>
         <card>
             <name>Graveborn</name>
-            <set picURL="http://i167.photobucket.com/albums/u148/Ruja/Magic%20Tokens/Coldsnap/5.jpg">CSP</set>
+            <set picURL="https://i.imgur.com/1PwSeto.jpg">CSP</set>
```
tooomm: New token is missing "haste"!

Psithief: That's consistent with Wizard's current token printing policy.

Psithief:
With tokens that haven't actually been printed by Wizards, there's often a limitation on how good these fakes/fan-arts are. Right now, reliability is more important.

tokens.xml
tooomm: They don't print `haste` any longer on tokens?
I guess that's the only keyword they don't print?

I must have missed that...

Psithief: When it is the only keyword on a token, they won't print it. This has been the case since at least Avacyn Restored, perhaps earlier.

PR #53, opened 2018-02-19
incorrect spelling

ZeldaZach:


Psithief:
Welp, time to add that as a check, finally.

Issue #52, opened 2017-12-17
Reminder: Delete "(Unknown card name - token)"


tooomm:
The latest, get rid of that when HOU rotates out of standard as stated in the code comment.

https://whatsinstandard.com/

Psithief:
This issue is a whole year early.

tooomm:
I'm sure I forgot about it in less time. ;P

Well, most likely the `(Unknown card name - token)` entry in the card and token list annoys me earlier...

Psithief:
Only 19 more days.

PR #57, opened 2018-10-03
remove unknown token

tooomm:
fixes https://github.com/Cockatrice/Magic-Token/issues/52

HOU rotates with the release of GRN at 5th October.


Psithief:
A couple hours of early! haha

PR #62, opened 2019-04-19
Version 4.

Psithief:
This doesn't function.

How many new properties are required?

Please review @ctrlaltca

### DIFF
--- a/tokens.xml
+++ b/tokens.xml
@@ -4,37 +4,48 @@
     We have made a new user-editable file for token modifications.
     Currently this file is "TK.xml" and can be found in your customsets directory.
 -->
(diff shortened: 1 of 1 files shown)


Psithief:
Changing `picURL` to `picurl` doesn't help.

Is this going to be something ridiculous like `maintype`, `layout` and `coloridentity` are mandatory?

ctrlaltca:
> This doesn't function.
What doesn't work exactly?

I've quickly coded a database converter tool that is able to load a v3 card database and save the v4:
https://github.com/Cockatrice/Cockatrice/pull/3694

ctrlaltca:
Attached is a converted version of the current v3 tokens file produced using the dbconverter tool.

[tokens.xml.txt](https://github.com/Cockatrice/Magic-Token/files/3098697/tokens.xml.txt)

Psithief:
The `sets` element is mandatory?
`maintype` is mandatory?

Psithief:
Apparently my version of Cockatrice didn't update all the way with just one update/install cycle.

Issue #66, opened 2019-07-16
Ajani's Pridemate token issue


ctrlaltca:
The token `Ajani's Pridemate` is currently overwriting the actual card named the same.
Original issue: https://github.com/Cockatrice/Cockatrice/issues/3784

PR #67, opened 2019-07-16
Make Ajani's Pridemade token unique

tooomm:
fix https://github.com/Cockatrice/Magic-Token/issues/66

Problem was that the actual card with the same name wasn't displayed anymore.
Used the same workaround we have for several instances of different Soldier tokens for example by appending space characters to the name.


Psithief:
@ZeldaZach Why are you merging these?

PR #69, opened 2019-07-18
Revert "Make Ajani's Pridemade token unique"

Psithief:
This is not the solution previously used.


tooomm:
You are right! I forgot about the special AKH creature tokens.

Psithief:
The originals were from Future Sight.

PR #72, opened 2019-07-28
Prettify and unify token names for card copies

tooomm:
Have all tags in the same style and make the token names look nicer.


Psithief:
I'd prefer if your commit messages weren't written all in lowercase.

PR #80, opened 2020-03-06
Updated Sliver tokens with correct sets

GuilhermoBarretti:
Corrected sliver token LGN to L13 and added sliver token P03 (Magic Player Rewards 2003)


ZeldaZach:
Thanks for the fix!

PR #82, opened 2020-03-29
Add info tag to tokens and create version file; fix #83

### LINKED ISSUE #83: Add version information to permit automated update checking
This issue is about adding the version informations to the tokens.xml file and to an external "version.txt" file.
The file could be edited manually or by some CI after a commit on master.
It should contain the same exact version string contained in the tokens.xml file to permit automatic checking of new available versions.
Please see https://github.com/Cockatrice/Magic-Token/pull/82 and https://github.com/Cockatrice/Cockatrice/pull/3934 for more info

ctrlaltca:
Please see https://github.com/Cockatrice/Cockatrice/pull/3934 for more info


tooomm:
Can we use a `version.json` file instead?
A text file looks not ideal for this kind of information.

Also, may I propose a different type for the version format?
Simply having a date as version leads to problems when there are several changes within one day for example, and it feels not like a proper version number.
But we could use the most current released expansion (number) and increment from there with new releases. Most changes happen on new releases.
Since it's difficult to keep track of all releases (special sets, some might not be available in all countries or don't have token etc.) I thought about only sticking to expansions as leading main version number. Unlike e.g. Core Sets they happen always and most likely don't stop. We want some reliable basis.
Theros Beyond Death is expansion number 83 (--> https://mtg.gamepedia.com/Theros_Beyond_Death), which would lead to `v83.0` as our starting point.
Little updates like fixes increase the following number, the same as a non-expansion set releases like commander, core set, un set or some duel decks and the like.
The next expansion release would bump the main version to 84 then. (Ikoria: Lair of Behemoths)

@Psithief What do you think? As you are mainly maintaining this.

ctrlaltca:
> Can we use a `version.json` file instead?

Yes, if you give me a good reason.

> A text file looks not ideal for this kind of information.

Why? Cockatrice only has to http GET it, read its content and see if the content differs from the one in the file it's actually using. 

> Also, may I propose a different type for the version format?

Of course: it's not forced to be a date, it can be any string.
I just used a date to get something in quickly. It just needs to change when we want to trigger the update.

Psithief:
I know how to write hooks for git but that will only automate the process for my own machine.

Isn't the proper way to do versioning an addition to the semantic versioning as in the [releases](https://github.com/Cockatrice/Magic-Token/releases) but with a build timestamp on the end?

e.g. `v.2.4.2b20200331123456`?

skwerlman:
the semver way would be to add a + followed by the metadata, like `v2.4.2+20200331123456`
see https://semver.org/#spec-item-10

tho if trice only needs to know if the value differs, cant you just use the git hash instead?

Psithief:
That pushes the burden to Cockatrice, and yes that makes it easier given there is no actual code in this repository.

Thinking about it, I don't see the benefit of forcing this repository to have continuous integration when there's nothing to compile. It's not that practical for contributors.

ctrlaltca:
Using CI to change the version.txt file automatically would trigger an update for all cockatrice users every single commit is merged on this repo. Maybe it's a better idea to control it manually and only trigger the update by modifying the contents of version.txt manually when needed.

Psithief:
It just seems extremely odd to me to duplicate this information in the repository contents.

Github's Releases are made _after_ the commit is created, so trying to put a version number for a release in the commit itself is just completely backwards.

Maybe we should have a new repository for a service that provides an answer to a version enquiry?
It would return the latest release version number through the Github API like @skwerlman mentioned.

Psithief:
I'd suggest the service could even insert the version element to files downloaded, but that seems like a waste of bandwidth costs when Cockatrice could do that instead.

ctrlaltca:
I don't get the need to complicate things around when a simple text file contaning a string would work without the need of commit hooks, CI deployments, semantic versioning, having to download json data from APIs and interpret them.

ctrlaltca:
I think I will just merge this now, so work on the new oracle version can continue.
This change won't affect anyone until the work will be finalized. In the meanwhile, we can discuss on #83 about the best way to implement this version information.

PR #89, opened 2020-07-16
README: add dynamic version badge

tooomm:
README now always shows the current version number

Preview: https://github.com/cockatrice/magic-token/tree/tooomm-patch-1#magic-token-

<br>

I'm still a fan of using a standard version number in semver style with its advantages ([and you use them for releases already...](https://github.com/cockatrice/magic-token/releases)). The date belongs to the currently empty `createdAt` key imo. In case that changes, I would display the date next to the version badge as well.


Psithief:
Fascinating stuff.

PR #90, opened 2020-07-24
Add token count badge to README

tooomm:
More badge magic... ever wanted to know how many tokens we serve?

Preview: https://github.com/Cockatrice/Magic-Token/blob/tooomm-patch-1/README.md#magic-token


Psithief:
This is not _as_ useful as the previous badge.

tooomm:
It's supposed to provide interesting general information about the scope/content of the project/file.

If you don't see use or fit and don't want to include it, close the PR please.
I came up with it for spoilers a while ago and thought it might be handy here as well... 👍 

Up to you. Just a suggestion. :)

Psithief:
I thought about it, but it's just not useful.

PR #91, opened 2020-08-27
add jumpstart

ebbit1q:
jumpstart doesn't have any tokens printed, this means the unicorn token
doesn't exist at all.


Psithief (approved):
Checks out.

PR #93, opened 2020-10-12
CI: Validate token XML files against Cockatrice's XSD schema

### LINKED ISSUE #65: have travis check xml file formatting and validade against xsd
Should have helped with the little errors we just corrected.

I guess the cockatrice xsd would have spotted the `<prop>` key issues?
But yeah, I know... no such updated file yet for v4: https://github.com/Cockatrice/Cockatrice/issues/3696

Travis setup is easy, but we would need e.g. a little python script to do the actual checks and produce some useful output in the travis logs.

tooomm:
fixes #65

Travis CI will now automatically check and report if tokens.xml is valid against Cockatrice DB schema v4, and challange-tokens.xml against Cockatrice DB schema v3.

On every change made to the repo or any PR created this will trigger and has results shortly after.

Checks are super fast, but it looks like we are not perfectly valid currently?


tooomm:
From the Travis logs:
```
$ xmllint --schema cards-v3.xsd --noout challenge_tokens.xml

challenge_tokens.xml:104: element card: Schemas validity error : Element 'card': Character content other than whitespace is not allowed because the content type is 'element-only'.

challenge_tokens.xml fails to validate

The command "xmllint --schema cards-v3.xsd --noout challenge_tokens.xml" exited with 3.
```

```
$ xmllint --schema cards-v4.xsd --noout tokens.xml

tokens.xml:7: element cockatrice_carddatabase: Schemas validity error : Element 'cockatrice_carddatabase', attribute '{https://www.w3.org/2001/XMLSchema-instance}noNamespaceSchemaLocation': The attribute '{https://www.w3.org/2001/XMLSchema-instance}noNamespaceSchemaLocation' is not allowed.

tokens.xml:1361: element card: Schemas validity error : Element 'card': Character content other than whitespace is not allowed because the content type is 'element-only'.

tokens.xml:5167: element prop: Schemas validity error : Element 'prop': This element is not expected. Expected is one of ( token, tablerow, cipt, upsidedown ).

tokens.xml:5602: element reverse-related: Schemas validity error : Element 'reverse-related': This element is not expected. Expected is one of ( tablerow, cipt, upsidedown ).

tokens.xml:7499: element set: Schemas validity error : Element 'set', attribute 'pickURL': The attribute 'pickURL' is not allowed.

tokens.xml:8291: element reverse-related: Schemas validity error : Element 'reverse-related': This element is not expected. Expected is one of ( tablerow, cipt, upsidedown ).

tokens.xml fails to validate

The command "xmllint --schema cards-v4.xsd --noout tokens.xml" exited with 3.
```

tooomm:
Tried to fix the obvious errors already in #94.
Also revealed a sneaky typo... I like this automated check already. 👍 

Please have a look at the spotted validation errors in this PR's and apply any further changes to #94 @Psithief!

Psithief:
As per https://github.com/Cockatrice/Cockatrice/pull/2787 and https://github.com/Cockatrice/Cockatrice/pull/3710

> The previously pushed XSD does not enforce at least one appearance any element that was required. It's not practical to implement this without a sequence with XML1.0 for this many elements.

>Unfortunately it's not perfect; the most problematic point is that some elements in our xml are not used as xml would like to:

> *   multiple set-per-card definitions are not nested inside a parent <sets> element;
> *   the same for card relations <related> and <reverse-related>;
>    This forces us into using <xs:sequence> instead of <xs:all>, that unfortunately also enforces the order of elements.

> These are mostly XSD 1.0 limitations that could be lifted by using an XSD 1.1 validator, but afaik no decent XSD 1.1 validator is easily available on every os/platform.

If we're automating checks, we really should try to revisit using an XML 1.1 XSD so we don't get complaints about a perfectly valid XML file.

Psithief:
> Checks are super fast, but it looks like we are not perfectly valid currently?

**No**, if it runs in Cockatrice, then the file is perfectly valid. The XSD is incomplete and incorrect documentation that defines a **subset** of the XML schema that will work with Cockatrice.

If we merge this, the checks are just going to be _ignored_ because they can't be trusted. Once that happens, you will find it difficult to get that trust back if a more accurate tool is implemented later.

tooomm:
Sure, that makes sense. As long as the checks are not helpful we should not merge this.

>If we're automating checks, we really should try to revisit using an XML 1.1 XSD so we don't get complaints about a perfectly valid XML file.

So the XSD at Cockatrice needs another update?

tooomm:
I revisited this PR to update it in order to use GitHub Actions over Travis to stay in line with all CI stuff in the Cockatrice organisation.

<br>

>The XSD is incomplete and incorrect documentation that defines a subset of the XML schema that will work with Cockatrice.

Had another look. While the error message by xmllint in CI looks like this:
>tokens.xml:7: element cockatrice_carddatabase: Schemas validity error : Element 'cockatrice_carddatabase', attribute '{https://www.w3.org/2001/XMLSchema-instance}noNamespaceSchemaLocation': The attribute '{https://www.w3.org/2001/XMLSchema-instance}noNamespaceSchemaLocation' is not allowed.

Using e.g. https://www.freeformatter.com/xml-validator-xsd.html, the error message is a bit more helpful:
>[Cvc-complex-type.3.2.2](https://wiki.xmldation.com/Support/Validator/cvc-complex-type-3-2-2): Attribute 'xsi:noNamespaceSchemaLocation' Is Not Allowed To Appear In Element 'cockatrice_carddatabase'., Line '6', Column '218'.

In the XSD, only the [`version` attribute](https://github.com/Cockatrice/Cockatrice/blob/master/doc/carddatabase_v4/cards.xsd#L85) is defined in the [`cockatrice_carddatabase` element](https://github.com/Cockatrice/Cockatrice/blob/master/doc/carddatabase_v4/cards.xsd#L57).
Do we need the [`xsi:noNamespaceSchemaLocation` attribute in the token.xml file](https://github.com/Cockatrice/Magic-Token/blob/master/tokens.xml#L7)? It got added in [this commit](https://github.com/Cockatrice/Magic-Token/commit/3a281817a57646112760809dd876411f393ff3c4#diff-ef2c5b2e0216513ba7314de2c6637a3b48b09a5ea3ee44d13c951195fb293697R6).
If yes, the XSD at Cockatrice should probably get updated to define it?

Somebody with better XML/XSD knowledge needs to help here. :)

<br>

This does not change Psi's point that this should not be merged until the XSD is updated and this check can be trusted. Which requires the XSD to be updated to 1.1.

Psithief:
> There is [xmlschema](https://pypi.org/project/xmlschema/), a Python library, with XSD 1.1 support!

Oh, great, how do we use that?

tooomm:
>Oh, great, how do we use that?

You can run Python in GitHub Actions CI, same as this PR runs xmllint currently.

In Cockatrice itself it would be a bit more tricky. But I'm not aware that Cockatrice utilizes the XSD files currently, so that doesn't matter I guess?

tooomm:
*Note to myself*

Problem matcher works nicely and annotates + adds link to the Action overview:
problem matcher

Also a great view on the changed files tab:

tooomm:
The mystical 

```
tokens.xml:7: element cockatrice_carddatabase: Schemas validity error : Element 'cockatrice_carddatabase', attribute '{https://www.w3.org/2001/XMLSchema-instance}noNamespaceSchemaLocation': The attribute '{https://www.w3.org/2001/XMLSchema-instance}noNamespaceSchemaLocation' is not allowed.
```

error was due to `xmlns:xsi` does not like https for some reason. [`c73fcaf` (#93)](https://github.com/Cockatrice/Magic-Token/pull/93/commits/c73fcaf43d8c309415f794f39ce9d633b0111258)

tooomm:
Useful example of the xmllint run:
#134 introduced an issue, which got spotted 4 month later solely by investigating an issue reported at Cockatrice, fixed in #150.

This CI run would have reported the issue in the introducing PR right away before even merging:
```
Error: tokens.xml:2104: element set: Schemas validity error : Element 'set', attribute 'picUrl': The attribute 'picUrl' is not allowed.
Error: tokens.xml:2692: element set: Schemas validity error : Element 'set', attribute 'picUrl': The attribute 'picUrl' is not allowed.
Error: tokens.xml:3681: element set: Schemas validity error : Element 'set', attribute 'picUrl': The attribute 'picUrl' is not allowed.
tokens.xml fails to validate
```

SlightlyCircuitous:
> The XSD is incomplete and incorrect documentation that defines a subset of the XML schema that will work with Cockatrice.
> Somebody with better XML/XSD knowledge needs to help here. :)

My understanding is that our cards.xsd files care about the order of elements in an entry, whereas Cockatrice doesn't really care as long as they are present.  For example, https://github.com/Cockatrice/Magic-Token/pull/96 rearranged two lines to make the validator happy, but Cockatrice can display all that information correctly either way. Based on Psithief's explanation, this seems to be necessary in our case because of the constraints of XSD 1.0. I agree that if this validator flagged element order issues frequently, it would get annoying and soon be ignored since the file is valid in Cockatrice. 

However, this seems like it would be a pretty rare false positive nowadays as we basically always put elements in the same order. cards.xml is generated automatically (right?), my updates to tokens.xml are generated near enough to automatically (and it seems like other people copy and paste the blank template if they do it manually so order is still preserved), and the challenge tokens will probably never be updated again (besides the proposition to swap them to v4, which was also done automatically). 

It sounds like it would definitely be advantageous to switch to XSD 1.1, but it doesn't seem like not being able to make that switch should block automatic XML validation. Though maybe I am missing some other important issues with XSD 1.0 since I am pretty new to this project.

PR #94, opened 2020-10-12
Fix validation errors

tooomm:
As spotted in https://github.com/Cockatrice/Magic-Token/pull/93

### DIFF
--- a/tokens.xml
+++ b/tokens.xml
@@ -5163,7 +5163,6 @@ Creatures you control attach each combat if able.</text>
         </card>
         <card>
             <name>Ragavan</name>
-            <set picURL="https://img.scryfall.com/cards/large/front/1/e/1ebc91a9-23e0-4ca1-bc6d-e710ad2efb31.jpg">AER</set>
             <prop>
                 <colors>R</colors>
                 <cmc>0</cmc>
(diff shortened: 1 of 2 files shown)


tooomm:
Leftover warning should only be

```
tokens.xml:7: element cockatrice_carddatabase: Schemas validity error : Element 'cockatrice_carddatabase', attribute '{https://www.w3.org/2001/XMLSchema-instance}noNamespaceSchemaLocation': The attribute '{https://www.w3.org/2001/XMLSchema-instance}noNamespaceSchemaLocation' is not allowed.
```

tooomm:
Too fast Zach, hoped that @Psithief has an idea about the other error. :P

(Version bump was also missing)

Psithief:
Those fixes are fine but why did the `<token>` field change lines?

tokens.xml
```diff
@@ -5598,9 +5598,9 @@ Creatures you control attach each combat if able.</text>
                 <pt>2/2</pt>
             </prop>
             <set picURL="https://media.wizards.com/2019/mh1/en_tM8ljjBCro.png">MH1</set>
-            <token>1</token>
```
tooomm, line 5601: >Those fixes are fine but why did the <token> field change lines?

Because the linter complained:
https://travis-ci.com/github/Cockatrice/Magic-Token/builds/189491625#L188

PR #97, opened 2020-10-14
README: add community+chat options

tooomm:
Include options to connect with us and report issues easier.


Psithief:
@tooomm Why would you commit this?
@ZeldaZach Why would you merge this?

Is it relevant or useful? I don't believe so.

tooomm:
In all Cockatrice repos we have this kind of links. In case people come here, they can easily connect with us and report issues without having GitHub accounts for example.

PR #100, opened 2021-02-23
Fix Monkey Cage Ape token

tooomm:
[Monkey Cage](https://scryfall.com/card/mmq/307/monkey-cage)'s Ape token was listed as Monkey.

I spotted this randomly when looking at https://www.mtg.onl/mtg-missing-tokens/.
There might be more.


Psithief:
You're incorrect. Look at the link you posted. The card text says ape but the **oracle** text says monkey.

tooomm:
Absolutely right. Subtle difference that matters. :)

PR #101, opened 2021-02-23
Remove x=0 values

tooomm:
It looks like Cockatrice always preselects a value of "1" in the `Create x tokens` dialog, even if "x=0" is set in the token file.

This PR removes all 7 occurrences  of `<reverse-related count="x=0">`.


Psithief:
Cockatrice used to give a zero option, why did it change?

tooomm:
Honestly, I can not tell if it used to be different.
But retested and confirmed the behavior again with `Client Version: 2.8.1-custom(efc3f98) (2021-02-09)` and [Scute Swarm](https://github.com/Cockatrice/Magic-Token/blob/master/tokens.xml#L3944).
The dialog doesn't allow you to enter "0" at all, it jumps right back to "1". Might be a Qt thing as well?

Maybe @ebbit1q has a clue?

ebbit1q:
> Honestly, I can not tell if it used to be different.
> But retested and confirmed the behavior again with `Client Version: 2.8.1-custom(efc3f98) (2021-02-09)`, latest token.xml file and [Scute Swarm](https://github.com/Cockatrice/Magic-Token/blob/master/tokens.xml#L3944).
> The dialog doesn't allow you to enter "0" at all, it jumps right back to "1". Might be a Qt thing as well?
> 
> Maybe @ebbit1q has a clue?

making 0 tokens is the same as doing nothing, it doesn't makes sense to allow people to choose x=0 so we intentionally set the dialog to have a range 1-99

also testing scute swarm, why does it have two entries?

tokens.xml
```diff
@@ -3941,9 +3941,9 @@ When this creature enters the battlefield, it deals 1 damage to any target.</tex
             <reverse-related count="5">One Dozen Eyes</reverse-related>
             <reverse-related>Saber Ants</reverse-related>
             <reverse-related>Scute Swarm</reverse-related>
```
ebbit1q, line 3943: I don't think we need this twice

tooomm: Since x defaults to 1, I do agree!

tooomm:
>making 0 tokens is the same as doing nothing, it doesn't makes sense to allow people to choose x=0 so we intentionally set the dialog to have a range 1-99

👍
I can not tell if that used to be different as Psi is recalling, but it makes sense that it works like that now.

All other `x=0` cases look ok. One could argue [Swarm Shambler](https://scryfall.com/card/znr/207/swarm-shambler) is not really an X case and there are only very few spells that target more than one and the targets all need to have counters too...

Psithief:
The reason I used `x=0` was to reduce input error for some cards. It wasn't used as consistently as perhaps it could have been. Now the feature is removed that's no longer possible.

The extra entry for Scute Swarm is intentional. 99% of the time, the player is only putting one land onto the battlefield at a time, but there are effects that will put multiple lands into play at once. This means that the default "All Tokens" keyboard shortcut creates one insect, but the player can use the context menu if they have six or more lands.

Whether that's a useful 'feature', well, I haven't questioned the player base to find out.

There are many examples of extra entries in the file and a pull request should only be addressing _one_ issue, so please remove the line deletion. If it's going to change, then **all** of them must change.

tooomm:
Got your point. 👍 

Ours was that the single `x` case covers all cases while still defaulting to the most needed x=1 case. Additionally the tokens.xml file and right click options in the UI would be more lightweight with no substantial drawback, but sure. The default shortcut still works correctly in both cases.

>but the player can use the context menu **if they have six or more lands**.

Note that if they have 6 or more lands, they create no more Insect tokens, but tokens that are copies of [Scute Swarm](https://scryfall.com/card/znr/203/scute-swarm) itself. So the token creation doesn't help here at all, they need to use the copy feature anyway.

Psithief:
Sorry, I was sure I'd made a mistake after I wrote that.

I meant to say they can use the context menu if an effect brought several lands into play at once. However I agree that's a bit unnecessary for Scute Swarm most of the time, as there's unlikely to be many effects that do that before there are six lands on that player's side of the battlefield.

ebbit1q:
I updated the pr but it's still saying it has conflicts

ebbit1q:
making an empty commit fixed it...

PR #112, opened 2021-07-24
Fixes error which breaks the rest of the tokens.xml

mainman879:


tooomm (approved):
Well spotted 👍

tooomm:
CI would prevent such errors to happen and spot them before users are affected. 😉 
Not a task for humans.

ebbit1q:
why isn't this merged yet, this is a breaking issue

PR #113, opened 2021-07-28
Fixes incorrect image link for Tomb of Annihilation

mainman879:


ebbit1q:
...

we should automate the dumb version tag...

PR #114, opened 2021-08-06
MIC and MID Tokens

mainman879:


mainman879:
@Psithief @tooomm @ebbit1q if you guys could take a look at what I have so far I would appreciate it. It should have every token relation from MID but I may have missed something and not realized it yet. There are a lot of missing picURLs right now but I can add those very easily later on when they get posted.

tokens.xml
```diff
@@ -9837,6 +10019,62 @@ Whenever a card is put into your graveyard from anywhere, you may return it to y
...
         </card>
+        <card>
+            <name>Day</name>
+            <prop>
+                <type>State</type>
+                <maintype>State</maintype>
+            </prop>
+            <set picURL="https://c1.scryfall.com/file/scryfall-cards/large/front/9/c/9c0f7843-4cbb-4d0f-8887-ec823a9238da.jpg?1630606483">MID</set>
+            <related attach="attach">Night<related>
+            <reverse-related exclude="exclude"></reverse-related>
```
ebbit1q, line 10030: why are these empty?

tokens.xml
```diff
@@ -9837,6 +10019,62 @@ Whenever a card is put into your graveyard from anywhere, you may return it to y
...
             <tablerow>1</tablerow>
         </card>
+        <card>
+            <name>Day</name>
+            <prop>
+                <type>State</type>
+                <maintype>State</maintype>
+            </prop>
+            <set picURL="https://c1.scryfall.com/file/scryfall-cards/large/front/9/c/9c0f7843-4cbb-4d0f-8887-ec823a9238da.jpg?1630606483">MID</set>
+            <related attach="attach">Night<related>
```
ebbit1q, line 10029: I don't think we want attach here, rather just have both make the other side, then destroy the original

tokens.xml
mainman879: Leftover from when I was indecisive on whether to have Day related to Night or Night reverse-related to Day.

tokens.xml
mainman879: I was unsure on whether to have it attach or not. I'll let some others state what they think about it as well.

tooomm:
You are probably aware, still want to note it: Bird, Ooze, 2x Zombie are missing pictures from MID

tokens.xml
tooomm: Having a quick look here...
Wondering why you have "Day" `reverse-related` and ` related` to "Night", but "Night" has no related tags to "Day". 🤔

mainman879: If Day is reverse related and related to Night, Night does not need those tags. Both will be able to create the other. Adding the tags again to Night would be completely redundant.

ebbit1q: the question is whether or not we want to have the "night" token attach to "day" on creation, in a similar fashion to transformed cards (which would be inconsistent when the "night" side is created alone) or we could have the token simply create the other side and then have the user spend another action cleaning up "day" (I think some users might prefer this as it looks a bit cleaner)

also relevant: we talked about adding the option to hide the parent card when creating an attached card, this would appear as if the parent card has transformed into the attached token, this would require a minor addition to the protocol

tooomm: >If Day is reverse related and related to Night, Night does not need those tags. Both will be able to create the other. Adding the tags again to Night would be completely redundant.

Sure, let me rephrase the question:
Why don't you have "Night" reverse-related in the "Day" entry the same way as you have "Day" reverse-related in the "Night" entry already. That's how entries are normally structured in the file?
I just realized that you vary from the norm here and chose a different way of representing two very similar token.

mainman879: If we want them to be reverse related to each other instead of related and reverse related on day that's fine. It has zero impact on functionality.

tooomm: It should be uniform for Day and Night, one way or the other.

We have not a strict pattern setup yet for e.g. token that create other token (e.g. Emblems).
Normally, we reverse-connect token->card. For the token->token case it's mixed.

mainman879: I think we should work towards making it consistent then. There are currently about 9 or so token relations that use related instead of reverse related. A separate PR changing all of these to use reverse related could be good.

tooomm: >I think we should work towards making it consistent then. There are currently about 9 or so token relations that use related instead of reverse related. A separate PR changing all of these to use reverse related could be good.

I fully agree here. I try to open an issue to track that later.

The file is organically grown and since @Psithief was mostly editing it alone, there was never a logic description made or needed.

I did compile some basic things in a [first, simple documentation](https://github.com/Cockatrice/Magic-Token/wiki#documentation) to explain the features and possibilities. If there is more, please add it!
There is nothing about the structure and ordering yet. Or patterns like card->token, token->token.

mainman879:
@Psithief I can't request a specific reviewer (because im not a maintainer or whatever I think), but wanted to let you know that it should be finished if you want to review it as well.

tokens.xml
tooomm: Thanks for opening https://github.com/Cockatrice/Magic-Token/pull/119, we can continue further discussion there.

Back to topic now: 😃 

>the question is whether or not we want to have the "night" token attach to "day" on creation, in a similar fashion to transformed cards (which would be inconsistent when the "night" side is created alone) or we could have the token simply create the other side and then have the user spend another action cleaning up "day" (I think some users might prefer this as it looks a bit cleaner)
>
>also relevant: we talked before about adding the option to hide the parent card when creating an attached card, this would appear as if the parent card has transformed into the attached token, this would require a minor addition to the protocol

I want to mention https://github.com/Cockatrice/Cockatrice/issues/4081 here.
There was the idea to have something like `<reverse-related attach="transform">` and allow for an even smoother support of double faced cards via card "turning over" like it would be done on a real table.
This would be a great application for Day/Night as well.

tokens.xml
```diff
@@ -9837,6 +10037,60 @@ Whenever a card is put into your graveyard from anywhere, you may return it to y
...
+            <reverse-related exclude="exclude">Sunrise Cavalier</reverse-related>
+            <reverse-related exclude="exclude">Sunstreak Phoenix</reverse-related>
+            <reverse-related exclude="exclude">Suspicious Stowaway</reverse-related>
+            <reverse-related exclude="exclude">Tavern Ruffian</reverse-related>
+            <reverse-related exclude="exclude">The Celestus</reverse-related>
+            <reverse-related exclude="exclude">Tireless Hauler</reverse-related>
+            <reverse-related exclude="exclude">Tovolar, Dire Overlord</reverse-related>
+            <reverse-related exclude="exclude">Tovolar's Huntmaster</reverse-related>
+            <reverse-related exclude="exclude">Vadrik, Astral Archmage</reverse-related>
+            <reverse-related exclude="exclude">Village Watch</reverse-related>
```
tooomm, line 10078: Hmm, why did you put the exclude tag for most of those Daybound cards?

<br>

There are also png pictures at scryfall: https://c1.scryfall.com/file/scryfall-cards/png/front/9/c/9c0f7843-4cbb-4d0f-8887-ec823a9238da.png?1630606483

I did not test it but I assume the jpg looks not as great with its white corners.
Wizards images looks to be png, scryfall mostly jpg in the rest of the file...

tokens.xml
```diff
@@ -9837,6 +10037,60 @@ Whenever a card is put into your graveyard from anywhere, you may return it to y
...
+        </card>
+        <card>
+            <name>Night</name>
+            <prop>
+                <type>State</type>
+                <maintype>State</maintype>
+            </prop>
+            <set picURL="https://c1.scryfall.com/file/scryfall-cards/large/back/9/c/9c0f7843-4cbb-4d0f-8887-ec823a9238da.jpg?1630606483">MID</set>
+            <reverse-related>Day</reverse-related>
+            <reverse-related exclude="exclude">Unnatural Moonrise</reverse-related>
```
tooomm, line 10090: Same here.

tooomm (commented):
Had a look at all related tags and the image links.

ebbit1q (approved):
yup, this works fine

tokens.xml
```diff
@@ -8530,6 +8659,52 @@ Sacrifice this creature: This creature deals 1 damage to target creature or play
...
+                <colors>B</colors>
+                <type>Token Creature — Zombie</type>
+                <maintype>Creature</maintype>
+                <cmc>0</cmc>
+                <pt>2/2</pt>
+            </prop>
+            <set picURL="https://media.wizards.com/2021/mid/en_gQ1kYIpAvu.png">MID</set>
+            <reverse-related count="x">Crowded Crypt</reverse-related>
+            <reverse-related>Curse of the Restless Dead</reverse-related>
+            <reverse-related count="x" exclude="exclude">Curse of the Restless Dead</reverse-related>
```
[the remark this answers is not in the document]
ebbit1q: 1 is the default for x

tokens.xml
```diff
@@ -1348,16 +1371,22 @@ Cloud Sprite can block only creatures with flying.</text>
...
+            <reverse-related>Briarbridge Tracker</reverse-related>
             <reverse-related>Bygone Bishop</reverse-related>
             <reverse-related>Byway Courier</reverse-related>
             <reverse-related>Confirm Suspicions</reverse-related>
             <reverse-related>Confront the Unknown</reverse-related>
             <reverse-related>Declaration in Stone</reverse-related>
+            <reverse-related>Dennick, Pious Apparition</reverse-related>
             <reverse-related>Drownyard Explorers</reverse-related>
+            <reverse-related>Eloise, Nephalia Sleuth</reverse-related>
+            <reverse-related count="x" exclude="exclude">Eloise, Nephalia Sleuth</reverse-related>
```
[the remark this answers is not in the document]
ebbit1q: 1 is the default for x

tokens.xml
```diff
@@ -3864,26 +3912,36 @@ At the beginning of your upkeep, this creature deals 1 damage to you.</text>
...
             </prop>
+            <set picURL="https://media.wizards.com/2021/mid/en_uHB1zQVvcu.png">MID</set>
             <set picURL="https://media.wizards.com/2020/iko/en_kkMVLXPjkL.png">C20</set>
             <set picURL="https://media.wizards.com/2019/eld/en_OnoAHjNCtA.png">ELD</set>
             <set picURL="https://media.wizards.com/2019/c19/en_0GA7rbmHav.png">C19</set>
             <set picURL="https://media.wizards.com/2018/rna/en_IeLQUTgkMy.png">RNA</set>
             <set picURL="https://img.scryfall.com/cards/large/front/8/8/8894949b-f190-461e-996a-cf2b39f08a5d.jpg">AVR</set>
             <set picURL="https://img.scryfall.com/cards/large/front/3/a/3a98c51f-c55c-4bd9-853b-65b8bad5a29f.jpg">FNM</set>
             <set picURL="https://img.scryfall.com/cards/large/front/b/d/bd5cd362-34f9-445f-85e1-9f6694f0f90a.jpg">DKA</set>
+            <reverse-related count="x">Adeline, Resplendent Cathar</reverse-related>
```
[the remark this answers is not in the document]
ebbit1q: x=1 is not useful

tokens.xml
ebbit1q: they all have white corners, cockatrice intentionally adds a black border to all cards to sort of counteract this

ebbit1q: the exclude tag means they don't get created together with normal tokens, this is what we want for states

tokens.xml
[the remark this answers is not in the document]
tooomm: We have many other `count="x=1"` occurrences in the file though. So much mixed stuff :)

I guess it's better practice to include it in case something changes at Cockatice's end?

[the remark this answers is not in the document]
mainman879: For previous cases where it was used we have: "Maja, Bretagard Protector", Hullbreacher, "Malcolm, Keen-Eyed Navigator", Death Tyrant, From Under the Floorboards, Ayula's Influence, and probably more as well. This is not a new format for it.

tokens.xml
mainman879: Everything being reverse-related is basically just because of me copy pasting the same format for all of them. Some of them should definitely be excluded because they make actual tokens besides Day.

tooomm: >they all have white corners, cockatrice intentionally adds a black border to all cards to sort of counteract this

Thanks for clarifying 👍 

>the exclude tag means they don't get created together with normal tokens, this is what we want for states

Sure, but none of those do create other tokens... hence the question.
There is mostly only Arlinn and Tovolar's Huntmaster doing so: https://scryfall.com/search?q=e%3Amid+o%3Adaybound&unique=cards&as=grid&order=name

tokens.xml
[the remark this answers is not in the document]
tooomm: >This is not a new format for it. (Using a regular non-exclude for 1 token, and then an excluded x token relation.)

For my understanding those double entries were meant for cards that have two particular cases/options (not just a ranging x value), e.g. [Chancellor of the Forge](https://scryfall.com/card/nph/81/chancellor-of-the-forge):
- Create 1/1 Goblin
- Create X 1/1 Goblins

(He has no exclude btw... should that only be set for recurring token creators? 🤔)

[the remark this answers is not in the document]
mainman879: The reason I did it like this is because for the vast majority of cases they will only create one token, and we should have this as default for ease of use. The x option is just there for instances where they wont create one token.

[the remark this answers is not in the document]
tooomm: That makes sense @mainman879. I'm not questioning your reason, your intention was quite obvious when looking at the card. However, as ebbit noted, x defaults to 1 already.
There is still the benefit of a bit quicker token creation with one step less (no dialog). That can even be a good reason for double entries.

Don't get me wrong. I just try to find common rules that everybody is aware of to work together on this and maintain a good quality. Again, there are no guidelines yet.
At the end, it's the user who is annoyed that it works different for similar cards.

tokens.xml
tooomm, line 10078:
```suggestion
            <reverse-related exclude="exclude">Arlinn, the Pack's Hope</reverse-related>
            <reverse-related>Baneblade Scoundrel</reverse-related>
            <reverse-related>Bird Admirer</reverse-related>
            <reverse-related>Brimstone Vandal</reverse-related>
            <reverse-related>Brutal Cathar</reverse-related>
            <reverse-related>Burly Breaker</reverse-related>
            <reverse-related>Celestus Sanctifier</reverse-related>
            <reverse-related>Component Collector</reverse-related>
            <reverse-related>Curse of Leeches</reverse-related>
            <reverse-related>Fangblade Brigand</reverse-related>
            <reverse-related>Firmament Sage</reverse-related>
            <reverse-related>Gavony Dawnguard</reverse-related>
            <reverse-related>Graveyard Trespasser</reverse-related>
            <reverse-related>Harvesttide Infiltrator</reverse-related>
            <reverse-related>Hound Tamer</reverse-related>
            <reverse-related>Kessig Naturalist</reverse-related>
            <reverse-related>Night</reverse-related>
            <reverse-related>Obsessed Astronomer</reverse-related>
            <reverse-related>Outland Liberator</reverse-related>
            <reverse-related>Reckless Stormseeker</reverse-related>
            <reverse-related>Shady Traveler</reverse-related>
            <reverse-related>Spellrune Painter</reverse-related>
            <reverse-related>Sunrise Cavalier</reverse-related>
            <reverse-related>Sunstreak Phoenix</reverse-related>
            <reverse-related>Suspicious Stowaway</reverse-related>
            <reverse-related>Tavern Ruffian</reverse-related>
            <reverse-related>The Celestus</reverse-related>
            <reverse-related>Tireless Hauler</reverse-related>
            <reverse-related>Tovolar, Dire Overlord</reverse-related>
            <reverse-related exclude="exclude">Tovolar's Huntmaster</reverse-related>
            <reverse-related>Vadrik, Astral Archmage</reverse-related>
            <reverse-related>Village Watch</reverse-related>
```

>There is only Arlinn and Tovolar's Huntmaster doing so: https://scryfall.com/search?q=e%3Amid+o%3Adaybound

Please double check.

tokens.xml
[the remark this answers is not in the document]
tooomm: >x=1 is not useful

"not useful"?

Are you referring to the default value that is set by Cockatrice and happens to be 1 as well?
Therefore `count="x"` and `count="x=1"` having the same result in the end?

Psithief (changes requested):
1. `Blessed Defiance` should link to a 1/1 white Spirit. Without the relation, how will the player access the token?

tokens.xml
```diff
@@ -9837,6 +10037,60 @@ Whenever a card is put into your graveyard from anywhere, you may return it to y
             <token>1</token>
             <tablerow>1</tablerow>
         </card>
+        <card>
+            <name>Day</name>
+            <prop>
+                <type>State</type>
+                <maintype>State</maintype>
```
Psithief, line 10044: Please add the text printed on the Day token. Italicised text is to be surrounded by parentheses like so:

```
<text>(If it becomes day or night or a daybound permanent enters the battlefield, track day/night for the rest of the game.)
As it becomes day, transform all nightbound permanents.
If a player casts no spells during their own turn, it becomes night next turn.</text>
```

tokens.xml
```diff
@@ -9837,6 +10037,60 @@ Whenever a card is put into your graveyard from anywhere, you may return it to y
...
+            <reverse-related exclude="exclude">Tovolar's Huntmaster</reverse-related>
+            <reverse-related exclude="exclude">Vadrik, Astral Archmage</reverse-related>
+            <reverse-related exclude="exclude">Village Watch</reverse-related>
+            <token>1</token>
+            <tablerow>1</tablerow>
+        </card>
+        <card>
+            <name>Night</name>
+            <prop>
+                <type>State</type>
```
Psithief, line 10085: Please add the text printed on the Night token.

```
<text>As it becomes night, transform all daybound permanents.
Permanents enter the battlefield nightbound.
If a player casts at least two spells during their own turn, it becomes day next turn.</text>
```

Psithief (changes requested):
Add <text> to day and night.

tokens.xml
```diff
@@ -8530,6 +8659,52 @@ Sacrifice this creature: This creature deals 1 damage to target creature or play
...
+            <reverse-related count="x">Ghoulcaller's Harvest</reverse-related>
+            <reverse-related>Ghoulish Procession</reverse-related>
+            <reverse-related>Hobbling Zombie</reverse-related>
+            <reverse-related>Jadar, Ghoulcaller of Nephalia</reverse-related>
+            <reverse-related>No Way Out</reverse-related>
+            <reverse-related>Poppet Stitcher</reverse-related>
+            <reverse-related>Revenge of the Drowned</reverse-related>
+            <reverse-related>Rotten Reunion</reverse-related>
+            <reverse-related>Startle</reverse-related>
+            <reverse-related count="x">Tainted Adversary</reverse-related>
```
Psithief, line 8688: This card can only create even numbers of tokens, so the minimum here should be 2, not 1.

tokens.xml
```diff
@@ -2548,6 +2581,21 @@ When Earthshaker Khenra enters the battlefield, target creature with power less
             <token>1</token>
             <tablerow>2</tablerow>
         </card>
+        <card>
+            <name>Elemental                            </name>
+            <text>This creature's power and toughness are each equal to the number of instant and sorcery cards in your graveyard plus the number of cards with flashback you own in exile.</text>
```
Psithief, line 2586: The text here is missing `Trample`.

Psithief (changes requested):

tokens.xml
```diff
@@ -7769,6 +7878,21 @@ Sacrifice this creature: This creature deals 1 damage to target creature or play
             <token>1</token>
             <tablerow>2</tablerow>
         </card>
+        <card>
+            <name>Vampire     </name>
+            <text>Trample, Lifelink, Haste</text>
```
Psithief, line 7883: Only the first keyword should be capitalised, to matched the printed token text.

Psithief (approved):

tokens.xml
mainman879: Look's right to me.

tokens.xml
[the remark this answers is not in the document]
mainman879: x and x=1 would result in the same exact thing yes. The only reason to use x=1 is if you expect cockatrice to change the default value of x away from 1.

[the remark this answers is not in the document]
tooomm: Ok, so nothing new then.

I still suggest to write that value to make sure it's always right and we don't rely on changes in the Cockatrice app or a new Qt version treating things differently.

There were already talks in other PR's whether x used to be 0 at one point. @Psithief raised that statement

tooomm:
Thanks for your review and potentially finding the last missing pieces, @Psithief. :)

There are still cases from discussions above, that we are unsure about - I wouldn't mind your opinion:
- Double entries for certain tokens from one card:
    - Having a dedicated "1" case because it often occurs next to its "x" case
    - Or combining it in a single "x" range with the default at "1"
- `count="x=1"`: When to use it and when not?
- `exclude`: How to do it for cards like [Chancellor of the Forge](https://scryfall.com/card/nph/81/chancellor-of-the-forge)

PR #116, opened 2021-08-22
Fix for nonexistent set

mainman879:
PPRE does not exist, and this angel token should refer to set P04 as seen here https://scryfall.com/card/p04/2/angel


ebbit1q (approved):
thanks

tooomm:
Ha, nice catch!

This fixes an issue mentioned in https://github.com/Cockatrice/Cockatrice/issues/4393.

PR #118, opened 2021-09-17
Remove all <related> for <reverse-related>

mainman879:
<related> was used a handful of times and never seemed to be very consistent when it was used. This change makes it so that every single relation uses <reverse-related> instead.


mainman879:
Didn't realize renaming a branch would close related PRs. That's a weird decision github.

Issue #111, opened 2021-07-11
Improve documentation


tooomm:
Expand on file documentation in the wiki and create and curate a new page.
Also include info about structure, layout and sorting.

The example card comment is also probably better off in the wiki next to its explanation and further details, and only have a link to all doc resource in the token file instead:
https://github.com/Cockatrice/Magic-Token/blob/master/tokens.xml#L15-L32

Helpful infos and hints:
- https://github.com/Cockatrice/Magic-Token/commit/a3838d95b259b3044fd6afdc6d536183fee89168
- [`67a2d6c` (#108)](https://github.com/Cockatrice/Magic-Token/pull/108/commits/67a2d6cb09cd3f1d4e4e03b9dc285455b33de36e)

ebbit1q:
I suggested adding a github action that simply automates all of this simple formatting stuff, ordering things is something computers should do, not humans

tooomm:
>ordering things is something computers should do, not humans

I totally support that!
Mentioned some potential other ideas in the linked issue as well. Most noticeable the ability to check if the date/version got updated, which is easily forgot. :)

There should probably still be some docs about the rules being applied/wanted.

Would you rather have a checking GH Action that notifies about misalignments that the PR creator should take care of with a script we provide (like the Cockatrice lint action?), or one that actually applies the preferred sorting rules and commits the changes right to the pr/repo (or probably better: creates a PR so that we can check again)?

The last (commit to repo) could run e.g. on push to master, scheduled once every week or triggered manually...

Psithief:
I don't really think the version should be changed except by the person merging a branch into master. It's not a useful request, because the author doesn't know when or if their pull request will be merged.

tooomm:
>the author doesn't know when or if their pull request will be merged.

True. That is a totally valid point!

>I don't really think the version should be changed except by the person merging a branch into master.

I agree with you on the time when such an update should happen.

There are ways around enforcing a manual human action there and skip the potential mistakes involved or forgetting it altogether.
As noted, the CI can run on every push events on master and update the date automatically the minute after. Or check e.g. on a daily schedule if changes happened and update the date with the last commit date etc.

ebbit1q:
actions can be triggered by merging

PR #120, opened 2021-09-22
Add documentation file

tooomm:
- [x] Remove example card from data file
- [x] Create new documentation file
    - [ ] Add dummy card
    - [ ] Add explanation about file structure and how relations work
    - [ ] Add explanation about various tags/attributes (include examples!)
    - [ ] Usage guidelines and best practices
- [ ] Link new file from current xml's
- [ ] Link to documentation in readme
- [ ] Remove notes from the wiki and link to the dedicated doc file instead

<br />

Useful resources:
- https://github.com/Cockatrice/Magic-Token/wiki#documentation
- https://github.com/Cockatrice/Cockatrice/wiki/Custom-Cards-&-Sets#to-add-your-own-custom-cards-follow-these-steps
- https://github.com/Cockatrice/Magic-Token/pull/119#issuecomment-922989535
- https://github.com/Cockatrice/Magic-Token/issues/111

### DIFF
--- a/documentation.md
+++ b/documentation.md
@@ -1,0 +1,1 @@
+Nothing here yet.
--- a/tokens.xml
+++ b/tokens.xml
@@ -12,24 +12,6 @@
         <sourceVersion>20210824</sourceVersion>
     </info>
     <cards>
-        <!--
-        <card>
-            <name></name>
-            <text></text>
-            <prop>
-                <colors></colors>
-                <manacost></manacost>
-                <type>Token Creature — </type>
-                <maintype></maintype>
-                <cmc>0</cmc>
-                <pt>/</pt>
-            </prop>
-            <set picURL=""></set>
-            <reverse-related></reverse-related>
-            <token>1</token>
-            <tablerow>2</tablerow>
-        </card>
-         -->
         <card>
             <name>Adorned Pouncer (Token)</name>
             <text>Double strike</text>


Psithief:
But I copy that when inserting tokens. 😢

mainman879:
I see no reason to delete it from the file. Having it in the file is easier to use than having it on a dedicated wiki page, and people may reference the token files when learning to make their own custom xmls.

tooomm:
>But I copy that when inserting tokens. 😢

I see, we are affecting and challenging the way you used to work with the file in many ways recently. Nothing personal. 🙈 

<br>

>I see no reason to delete it from the file. Having it in the file is easier to use than having it on a dedicated wiki page, and people may reference the token files when learning to make their own custom xmls.

In my opinion, a documentation should not be part of such a big file itself. Especially since it needs more than this dummy card. We can just not provide that in the file.

I know it's a handy helper and it should be noted somewhere (for e.g. copying) for sure! But does it really matter where one copies from?

Having a link in the file to a comprehensive documentation is more helpful and cleaner for everybody in the end, no?

mainman879:
> Having a link in the file to a comprehensive documentation is more helpful and cleaner for everybody in the end, no?

Is there a link for this documentation in the file itself already? If not, this dummy token could be replaced with a link I suppose.

tooomm:
No link yet, no. This should not be merged as is, as long as we are not decided of course.

<br>

I thought about gathering all useful information including this dummy in a new `Documentation` subpage on the wiki.

If you guys prefer a markdown file in the repo itself that's also fine! Maybe even better and easier to find!

Psithief:
> If you guys prefer a markdown file in the repo itself, that's also fine! Maybe even better and easier to find!

That would be far more convenient than a link.

tooomm:
I would still link such a `documentation.md` file from our xml for ease of use.

e.g. [`https://github.com/Cockatrice/Magic-Token/blob/master/README.md`](https://github.com/Cockatrice/Magic-Token/blob/master/README.md) (but a new file of course)

PR #122, opened 2021-09-22
Cleanup erroneous relationships

tooomm:
It was not just Kiora that needed a fix. ~~There are probably still more.~~

This also removes double linkings from both directions for some tokens via `related` and `reverse-related`.

<br>

Linking https://github.com/Cockatrice/Magic-Token/commit/0aa71e3d538ce355fe1b1d7ca9539a3510d08e7e

### DIFF
--- a/tokens.xml
+++ b/tokens.xml
@@ -1052,7 +1052,7 @@ Creature tokens you control have flying and vigilance.</text>
             <set picURL="https://media.wizards.com/2020/iko/en_kfvZCnHFzR.png">IKO</set>
             <set picURL="https://img.scryfall.com/cards/large/front/b/3/b31f1580-5bba-4cef-b0c8-f2837a597b7d.jpg">M19</set>
             <set picURL="https://img.scryfall.com/cards/large/front/a/f/af0086cd-56b5-450e-818c-e54f45d1a104.jpg">AKH</set>
-            <reverse-related count="3" exclude="exclude">Ajani, Adversary of Tyrants</reverse-related>
+            <reverse-related count="3">Ajani, Adversary of Tyrants (Emblem)</reverse-related>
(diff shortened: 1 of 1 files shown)


Psithief:
I think perhaps there's more than one change here so please give me a bit to examine this.

tooomm:
>I think perhaps there's more than one change here so please give me a bit to examine this.

Absolutely, please always double check! 👍 

<br>

If you follow the commits one by one it should be straight forward:

1) Removed two erroneous walker relation (+fixed a token count along the way)
2) Cleaned up the now present double linking (removes the fixed token count from 1) completely)
3) Removed another erroneous walker relation

tokens.xml
```diff
@@ -8460,7 +8459,7 @@ Sacrifice this creature: This creature deals 1 damage to target creature or play
             <reverse-related>Liliana, Death's Majesty</reverse-related>
             <reverse-related>Liliana, Dreadhorde General</reverse-related>
             <reverse-related>Liliana, Heretical Healer</reverse-related>
-            <reverse-related count="x">Liliana, the Last Hope</reverse-related>
+            <reverse-related count="x">Liliana, the Last Hope (Emblem)</reverse-related>
```
Psithief, line 8462: The minimum number of zombies is 2, so what about `"x=2"` here?

Psithief:
The weird part is, I think I kept doing it because I thought the first time was on purpose. 
This will be good to clean up.

tokens.xml
tooomm: Sure, that works fine.

The x count and it's "correct usage" in various situations is still a bit obscure.
No consensus in the file or the people working with it. That makes it a bit tricky to find the "right" solution as there are often several ways to utilize it.

PR #123, opened 2021-09-22
Add missing BBD token

### LINKED ISSUE #55: Battlebond
Tokens were posted to twitter instead of wizards.com. This seems like an unreliable place to link to.

Update when there is a better location available.

tooomm:
Fixes #55


ebbit1q:
some of these are out of order aren't they?

tooomm:
> some of these are out of order aren't they?

Double checked them again, there was one mistake. 👍 https://github.com/Cockatrice/Magic-Token/commit/3b56dc1824648d4988cb454e4ab5c2170f7f5a97

PR #125, opened 2021-10-30
VOW Tokens

mainman879:


tokens.xml
```diff
@@ -1899,6 +1917,20 @@ When this creature dies, create a 2/2 red Dragon creature token with flying and
             <reverse-related>Nesting Dragon</reverse-related>
             <token>1</token>
             <tablerow>2</tablerow>
+        </card>
```
ebbit1q, line 1920:
```suggestion
        </card>
        <card>
```

mainman879:
Question for ordering cards when spaces are involved: Do we order it letter by letter (thereby ignoring the spaces), or word by word? The reason I ask is because of:

- <reverse-related>Bloodcrazed Socialite</reverse-related>
- <reverse-related>Bloodsoaked Reveler</reverse-related>
- <reverse-related>Bloodtithe Harvester</reverse-related>
- <reverse-related>Bloodvial Purveyor</reverse-related>
- <reverse-related>Blood Petal Celebrant</reverse-related>

The Blood Petal Celebrant has a space but the others dont, so if we go letter by letter it goes between bloodcrazed and bloodsoaked, and if we go word by word it goes before all of them.

ebbit1q:
space is a character, it is a lower value than others

tooomm:
To clarify on your statement, is A a lower value than B? Or the other way around?

mainman879:
Known issue: There is no token image for Timothar, Baron of Bats. It is a 1/1 Bat token with flying and another effect. I could not find an image of this token with the other tokens.

ebbit1q:
if no placeholder art can be found for the card you should remove the picURL alltogether

ebbit1q:
actually, timothar specifies that the card gains the ability, this means it is not part of the card's text box imo

PR #132, opened 2022-02-05
Add Mowu token

tooomm:
`Legendary 3/3 green Dog creature token` (named "Mowu"), created by [Jiang Yanggu](https://scryfall.com/card/gs1/22/jiang-yanggu), was missing.

Note, that in the print it still says `Hound`.


tokens.xml
```diff
@@ -5402,6 +5402,21 @@ A card with morph can be turned face up any time for its morph cost.)</text>
...
+            <name>Mowu</name>
+            <prop>
+                <colors>G</colors>
+                <cmc>0</cmc>
+                <type>Token Legendary Creature — Dog</type>
+                <maintype>Creature</maintype>
+                <pt>3/3</pt>
+            </prop>
+            <set picURL="https://c1.scryfall.com/file/scryfall-cards/large/front/b/1/b10441dd-9029-4f95-9566-d3771ebd36bd.jpg?1626729696">GS1</set>
+            <set picURL="https://c1.scryfall.com/file/scryfall-cards/png/back/b/1/b10441dd-9029-4f95-9566-d3771ebd36bd.png?1626729696">GS1</set>
```
ebbit1q, line 5415: why is one of these a jpg and the other a png? you know that one will overwrite the other and there is no reason to add both right?

tooomm: I'm not sure if we have a convention which one to favour actually. The jpg is smaller in size, the png probably better quality.
It's mixed in the rest of the file...
I simply copied the defaults from the scryfall page.

Check `image_uris` (https://api.scryfall.com/cards/search?q=mowu) to see all available ones for the front.

I like to see it uniform whenever possible.

<br>

>you know that one will overwrite the other and there is no reason to add both right?

Isn't only the first one used?
I added the second for completeness and in case somebody wants the use the back side manually.

ebbit1q: noone should manually edit the tokens file on the client as it gets overwritten, we should prefer large jpg like the client does imo

ebbit1q:
```suggestion
            <set picURL="https://c1.scryfall.com/file/scryfall-cards/large/back/b/1/b10441dd-9029-4f95-9566-d3771ebd36bd.jpg?1626729696">GS1</set>
```

PR #135, opened 2022-02-07
add workflow for updating the version on pr approval

ebbit1q:
adds a workflow that triggers when a reviewer accepts a pull request that will subsequently update the pr by merging master into it and then updating the version number, pushing into the branch owned by the contributor.
- only members of the organization will trigger the workflow
- the pull request needs to be editable ( this is on by default )
- the workflow checks if changes to tokens.xml are present first
- the workflow does not affect master and the pull request should be merged immediately after verification by the reviewer
- the workflow does not handle two separate prs getting accepted at the same time, organization members should accept a pr and then merge it before accepting another pr ( the only bad thing that happens is that the version number will not get updated twice but only once, which is something that happens regularly right now )


tooomm (commented):
- add `workflow_dispatch:`
I think having the ability to quickly and easily trigger a version update manually if need be is always useful.

- adjust PR trigger:
Since there is no "run on PR merge" option and using "push to master" (which includes merged PRs as well as quick manual fixups) would result in a loop because the newly commited date to master branch will trigger another run...
...you probably went with "run when a PR reviews is submitted" and made sure they are "approving" ones.

    I see some potential issues with this, for example:
    - It's required to approve and merge the PR right after, otherwise the version is updated on master already, but the actual data changes are still sitting in the PR
    - Potential conflicts because the date in the file is updated which changes the PR target before the merge etc.
    - Everybody submitting an approving PR review will trigger a change on master immediately, even if he can not merge the PR himself (or they want to wait for further feedback and approval)

    There is also another option to circumvent the mentioned issues:
Run on closed PRs. Similar to your approach, where you focus on approving reviews we need to make sure to only run it when the event was an actual merge, not a "close with unmerged commits". 

    We can also narrows it further down to:
    - Only run when xml files are changed (approving this PR right now, would update the version on master needlessly)
    - Only run when the PR is targeting master (otherwise there is no reason to change the version on master)

- I think it would be useful to have the logic in the script, and do the git push in a separate workflow step

- Do we need to worry about cancelling previous runs of this workflow, for example if there are several changes being merged in a short time frame?

.github/workflows/updateversion.sh
```diff
@@ -0,0 +1,53 @@
...
+echo "$date" >version.txt
+
+# push changes
+git config user.name github-actions
+git config user.email github-actions@github.com
+git add tokens.xml version.txt
+git commit -m "update version to $date"
+git push
+commit=$(git rev-parse HEAD)
+echo "created commit: $GITHUB_SERVER_URL/$GITHUB_REPOSITORY/commit/$commit"
```
tooomm, line 53:
```suggestion
```

.github/workflows/updateversion.yml
```diff
@@ -0,0 +1,17 @@
...
+  update_version:
+    if: github.event.review.state == 'approved'
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Update date
+        shell: bash
+        run: .github/workflows/updateversion.sh
```
tooomm, line 17:
```suggestion
      - name: Update date
        id: update_date
        shell: bash
        run: .github/workflows/updateversion.sh

      - name: Push new version info to master
        if: steps.update_date.outcome == 'success'
        shell: bash
        run: |
          git config user.name github-actions
          git config user.email github-actions@github.com
          git add tokens.xml version.txt
          git commit -m "update version to $date"
          git push
          commit=$(git rev-parse HEAD)
          echo "created commit: $GITHUB_SERVER_URL/$GITHUB_REPOSITORY/commit/$commit"
```

ebbit1q:
you misunderstood what the intent of the pr is, the new workflow is intended to change the pr itself before it is merged, I wanted to have the workflow also merge the pr but this doesn't seem to work out well I think

ebbit1q:
@tooomm I've split the thing up like you wanted, and added a manual trigger in the form of @github-actions update version

PR #139, opened 2022-03-27
change set of the serpent generator token to legends

ebbit1q:
for some reason this used to be med which probably wasn't even an existing set code when we added the token?


ebbit1q:
please approve first, then merge

ZeldaZach (approved):

ebbit1q:
:facepalm: this is not how the automation is supposed to work

PR #141, opened 2022-04-23
Streets of New Capenna tokens

Psithief:
Let me know if I've made any mistakes or haven't caught up on any recent decisions.

I was surprised to discover that Wizards is printing tokens with only "Haste" of them again.

I couldn't figure out how to get at the token images on Scryfall, so the tokens are from the article by Wizards.


tokens.xml
```diff
@@ -1352,6 +1370,32 @@ This creature can block only creatures with flying.</text>
...
+            <reverse-related attach="attach">Citizen's Crowbar</reverse-related>
+            <reverse-related>Courier's Briefcase</reverse-related>
+            <reverse-related>Darling of the Masses</reverse-related>
+            <reverse-related>Exhibition Magician</reverse-related>
+            <reverse-related>Halo Fountain</reverse-related>
+            <reverse-related>Luxurious Libation</reverse-related>
+            <reverse-related count="x">Rabble Rousing</reverse-related>
+            <reverse-related>Rakish Revelers</reverse-related>
+            <reverse-related>Sanctuary Warden</reverse-related>
+            <reverse-related count="x" exclude="exclude">Stimulus Package</reverse-related>
```
ebbit1q, line 1394: I don't think this is an x ability, just exclude imo

tokens.xml
```diff
@@ -5553,6 +5634,21 @@ A card with morph can be turned face up any time for its morph cost.)</text>
...
+            <text></text>
+            <prop>
+                <colors>B</colors>
+                <type>Token Creature — Ogre Warrior</type>
+                <maintype>Creature</maintype>
+                <cmc>0</cmc>
+                <pt>4/3</pt>
+            </prop>
+            <set picURL="https://media.wizards.com/2022/snc/en_KvyeDaaJ91.png">SNC</set>
+            <reverse-related>Join the Maestros</reverse-related>
```
ebbit1q, line 5648: would having an extra count="2" exclude="exclude" be worth having here?

tokens.xml
```diff
@@ -6325,6 +6438,21 @@ Equip {1}</text>
...
         </card>
+        <card>
+            <name>Rogue Token</name>
+            <text></text>
+            <prop>
+                <colors>B</colors>
+                <type>Token Creature — Rogue</type>
+                <maintype>Creature</maintype>
+                <cmc>0</cmc>
+                <pt>2/2</pt>
```
ebbit1q, line 6449: what's our policy on tapped tokens? do we add cipt?

tokens.xml
```diff
@@ -8087,6 +8230,11 @@ This creature can't be enchanted.</text>
                 <maintype>Artifact</maintype>
                 <cmc>0</cmc>
             </prop>
+            <set picURL="https://media.wizards.com/2022/snc/en_hLVcbOyxF3.png">SNC</set>
+            <set picURL="https://media.wizards.com/2022/snc/en_XHQoWUdbKk.png">SNC</set>
+            <set picURL="https://media.wizards.com/2022/snc/en_u1BnldwqLg.png">SNC</set>
+            <set picURL="https://media.wizards.com/2022/snc/en_5j7mwsCJD8.png">SNC</set>
+            <set picURL="https://media.wizards.com/2022/snc/en_AXaJsB76gI.png">SNC</set>
```
ebbit1q, line 8237: there still isn't any use adding multiple copies of a set tag

tokens.xml
ebbit1q: currently Ragavan is the one token using cipt, I think this is more suited to a cockatrice change instead of setting the token to cipt  https://github.com/Cockatrice/Cockatrice/issues/4573

ebbit1q:
I've added the suggestions I made in my comments for convenience, we can remove the extra lines if you disagree

tokens.xml
```diff
@@ -138,8 +138,8 @@
             </prop>
             <set picURL="https://media.wizards.com/2022/snc/en_kCYPFaz3D0.png">SNC</set>
             <set picURL="https://img.scryfall.com/cards/large/front/9/6/96f407ef-2b08-45e9-a45e-128cf0a63839.jpg">OGW</set>
-            <reverse-related>Linvala, the Preserver</reverse-related>
             <reverse-related count="5">Elspeth Resplendent</reverse-related>
```
Psithief, line 141: I don't even remember seeing Linvala there. :)

tokens.xml
Psithief: Yes, the only use for this is for someone to pull their favourite token URL out of the file and make it custom, and that's pretty unlikely.
But, you as a reviewer can say which one you'd like to keep. ;)

tokens.xml
Psithief: This one is almost not worth the effort. If someone has the default keybinding, they're still holding down two modifiers and could just press `T` twice.

It's more useful if the numbers have a gap, like `Gather the Townsfolk` (two or five), `Throne of Empires` (one or five), `Starnheim Unleashed` (one or X). 
(`Saproling Migration` (two or four) or `Conquerer's Pledge` (six or twelve) aren't as good as examples because they're exactly double.)

tokens.xml
Psithief: Yes, I was annoyed by this when I tested it.

tokens.xml
tooomm: I would like to keep them. That's what we do for other variants within the same set as well.
No harm in keeping them, and yes, some users like the additional info when they want to change it.

tokens.xml
tooomm:
```suggestion
```

I do agree, only exclude. 👍

tokens.xml
tooomm: I think it's fine to have both.

People are probably using less shortcuts as you might think and simply go with the mouse. :)

tooomm:
@ebbit1q Can you pour your sorting magic into a script that outputs a e.g. diff and we have CI run it?

Psithief:
I was about to ask the same thing.

tokens.xml
Psithief: Alright, but which one do the two of you like most? Which one gets to see the light of day? 😁

ebbit1q:
well, it's a vim macro, I could just have it run vim with the macro in ci but there are more elegant solutions

tokens.xml
tooomm: Isn't that the honor of the person putting the work in to add them to decide? 😉

My favourites are nr. 1 and nr. 5 in your list.

ebbit1q (approved):
I think this looks good, tested most of the relations as well

tooomm:
>couldn't figure out how to get at the token images on Scryfall, so the tokens are from the article by Wizards.

You can look at https://scryfall.com/sets and search for the set in question, there normally is a dedicated "token" subset.
Putting, e.g. `s:tsnc` in the search accomplishes the same. But currently, there is only 1 token available yet.

The image links themselves you just grab via right click from the details page.
There are also direct links on the buttom in the `Images and data` section.
You can can also pull them out from the API directly, e.g.
https://api.scryfall.com/sets/tsnc --> following `search_uri`: https://api.scryfall.com/cards/search?order=set&q=e%3Atsnc&unique=prints and looking at `image_uris:`

tokens.xml
ebbit1q: it picks the last one and I thought it was pretty good, so it stays

PR #142, opened 2022-04-28
set kaito's ninjas to be 1/1 instead of */*

ebbit1q:


ebbit1q:
@github-actions update version

Issue #140, opened 2022-04-22
Ophiomancer's Snake Token Art


tooomm:
>[Ophiomancer](https://scryfall.com/card/cc2/3/ophiomancer)'s Token should be updated with the official artwork, since the current artwork was a substitute because there wasn't an official one.

_Originally posted by @BatmenzDW in https://github.com/Cockatrice/Magic-Token/issues/85#issuecomment-1105794980_

(Friendly reminder: Please try to create separate issues for new problems)

tooomm:
The newly available, official art is from CC2, which is not included yet.

Issue #92, opened 2020-10-10
Scryfall image link changes


tooomm:
As announced in their blog: https://scryfall.com/blog/upcoming-api-changes-to-scryfall-image-uris-and-download-uris-221

>Starting in mid-August, Scryfall will be transitioning our image and data origins to new subdomains. Images and bulk data items will stop being hosted on `img.scryfall.com` and `archive.scryfall.com` and they will move to new federated hosts at `c1.scryfall.com`, `c2.scryfall.com`, `c3.scryfall.com`, and so on, with new file paths.
>
>**Why are you doing this?**
>
>Scryfall is transitioning to a new file hosting provider to improve our scalability and also save money.
>
>**Will this break my existing code?**
>
>Only if you’ve been naughty. We don’t recommend that you ever try to manually construct a link to a file on `img.scryfall.com` or `archive.scryfall.com`. Instead you should fetch card data using our API or bulk file offerings, those objects always include the correct and up-to-date URI to card images and downloadable files.
>(...)
>
>If you have already been using the API to retrieve image and file URIs, nothing should change for you.
>
>**What if I have cached the old links? Will you set up any redirects?**
>
>Yes, we will be creating redirects for `img.scryfall.com` and `archive.scryfall.com`, but you should not permanently rely on these redirects. They should be considered temporary band-aids until you are able to update anything affected by the domain changes.
>
>**Can I still hotlink to the new origins?**
>
>Yes.
>
>**Do the new origins have a rate limit?**
>
>No, only api.scryfall.com has a rate limit.

---

Other notes:
- Scryfall also has PNG and JPG pictures. We currently use both mixed across the file
See [`image_uris` for "Academic Probation"](https://api.scryfall.com/cards/search?q=Academic%20Probation) for example.
- There is always a number attached to each picture. Should we scrape that for a cleaner file?
https://c1.scryfall.com/file/scryfall-cards/large/front/0/5/05521edf-f47f-4e7a-aec5-cdc4ae7368c2.jpg?1624589179

<br>

List of all tokens available at scyrfall (per set): https://scryfall.com/sets?type=token

Psithief:
Oh, fun. 👎

tooomm:
@ebbit1q Is this something you can try to look into as well and find a nice automated solution maybe? 😉 

We could use API redirects, or go with the new file locations before they stop maintaining the old redirects.

tooomm:
Looks like `img.scryfall` links are finally 404ing.

[...]

PR #143, opened 2022-05-07
Fixes 2 snake tokens

mainman879:
fixes https://github.com/Cockatrice/Magic-Token/issues/140 Ophiomancer didn't have an official token until now so this adds that. The snake for Hapatra Vizier of Poisons had a broken link so this also fixes that.


tokens.xml
```diff
@@ -7058,7 +7058,7 @@ Equip {1}</text>
                 <cmc>0</cmc>
                 <pt>1/1</pt>
             </prop>
-            <set picURL="https://media-dominaria.cursecdn.com/attachments/113/426/635201279747658553.jpg">C13</set>
+            <set picURL="https://c1.scryfall.com/file/scryfall-cards/large/front/6/6/66d80dd1-b944-4cb2-8578-b4dbcabbbc1e.jpg?1562916501">C13</set>
```
ebbit1q, line 7061: this is a picture of Ophiomancer instead

tokens.xml
```diff
@@ -7073,7 +7073,7 @@ Equip {1}</text>
                 <cmc>0</cmc>
                 <pt>1/1</pt>
             </prop>
-            <set picURL="https://img.scryfall.com/cards/large/front/6/a/6aaa8539-8d21-4da1-8410-d4354078390f.jpg">AKH</set>
+            <set picURL="https://c1.scryfall.com/file/scryfall-cards/large/front/6/a/6aaa8539-8d21-4da1-8410-d4354078390f.jpg?1562844799">AKH</set>
```
ebbit1q, line 7076: this has been coming for a while but every single image that starts with img.scryfall is broken

ebbit1q: see https://github.com/Cockatrice/Magic-Token/issues/92

tokens.xml
mainman879: Whoops, had multiple JSONs open at once and forgot to double check which was which. Fixed.

tokens.xml
mainman879: It looks like it could be fixed with automation. 

Change img to c1 and add in file/scryfall- after com/

As far as I can tell the ?number at the end of the link affects nothing, and the link works the same without it from what I can see.

ebbit1q: we have had one situation where the ? actually did affect the shown picture in scryfall over on the spoilers repo, the spoilers repo used to trim it but we reverted that

tokens.xml
```diff
@@ -7058,7 +7058,7 @@ Equip {1}</text>
                 <cmc>0</cmc>
                 <pt>1/1</pt>
             </prop>
-            <set picURL="https://media-dominaria.cursecdn.com/attachments/113/426/635201279747658553.jpg">C13</set>
+            <set picURL="https://c1.scryfall.com/file/scryfall-cards/large/front/1/3/13e4832d-8530-4b85-b738-51d0c18f28ec.jpg?1630007268">C13</set>
```
ebbit1q, line 7061: this is a low quality scan, I think it's just the wizards token but upscaled

tooomm: That's all they have currently:
https://api.scryfall.com/cards/13e4832d-8530-4b85-b738-51d0c18f28ec

While being a bit bigger in resolution and way bigger in size... the png seems to be the same?

tokens.xml
tooomm: The official info from Scryfall support is that they use the integer to bust the CND caches for various reasons.
One should use the URI they provide with the integer parameter at the end to ensure one gets the best version of the image currently available.

Not so sure that is super relevant for us as we cache pictures locally anyway and the client never realizes that a parameter got updated to redownload the file and update the local version.

tokens.xml
ebbit1q: actually, after I've merged this, was this token actually printed in C13?

tooomm: It's printed in CC2 [as far as I'm aware](https://github.com/Cockatrice/Magic-Token/issues/140#issuecomment-1106369825).

PR #144, opened 2022-05-09
Fix set for snake token

tooomm:
See https://github.com/Cockatrice/Magic-Token/pull/143#discussion_r867586945

@ebbit1q How to bump the version number on my own PR as I can't approve my own changes?
I saw you mentioned the bot in #142? Can you please explain how to use this correctly as I think its not documented?

Edit: Ok, I see there are some special cases in the code. That also explains why the [workflow is triggered so many times](https://github.com/Cockatrice/Magic-Token/actions/workflows/updateversion.yml) (for every issue or pull request comment that is created actually). I put some notes down and adjusted this in #145.


mainman879:
I do not think it is a good idea to change the only existing token for this card away to CC2. Yes it was printed in CC2 but CC2 is not a main set, and is a different set from the token generator which is in C13. This would mean that if someone disabled any "non-core" sets this card wouldn't work for them anymore.

tooomm:
>I do not think it is a good idea to change the only existing token for this card away to CC2. Yes it was printed in CC2 but CC2 is not a main set, and is a different set from the token generator which is in C13.

Its token generator is in both sets.
I've no strong opinion here. Maybe @Psithief?

>This would mean that if someone disabled any "non-core" sets this card wouldn't work for them anymore.

The right click in a game and create related tokens feature should always work if I'm not mistaken?
It would not show up in the list of cards in the deck editor though.

mainman879:
> The right click in a game and create related tokens feature should always work if I'm not mistaken? It would not show up in the list of cards in the deck editor though.

I had just tested this by editing my own xml and it did work even after disabling CC2. I had assumed the program would completely prevent unselected sets from being used but I guess not. So changing it seems fine if it doesn't really matter.

ebbit1q:
no, disabling actually does this: it hides those cards from the deck editor and it puts the set last in order for art selection

ebbit1q:
@tooomm are there more tokens in CC2 that we should consider adding?

tooomm:
@ebbit1q There is a Zombie token on the back of the snake. But that's about it. #146

PR #145, opened 2022-05-09
CI: Adjust + document version bump workflow

tooomm:
Changes:
- Little doc about how this works.
- Do not trigger on every comment written in all repo, but only on *submitted reviews*.
    - In case the PR creator themselves need to trigger the workflow, they can submit a *review comment* on their own PR.
- Changed keywords to ❌ and 🚢

<br>

TODO
- [ ] Update https://github.com/Cockatrice/Magic-Token/wiki/Token-updating-process-flow

### DIFF
--- a/.github/workflows/updateversion.yml
+++ b/.github/workflows/updateversion.yml
@@ -1,11 +1,11 @@
-name: Update version on pr approval
+name: Update version on PR approval
 
 # Usage of this automated version bump workflow
 #
 # If there is a new PR with valid changes and additions, there are two ways to let the CI run before merging:
-#  a) Approve the changes via the PR review feature.
-#     In case you want to approve the PR, but need to skip the version update, let the bot know. ("@github-actions :x:")
-#  b) Since PR creators can not approve their own changes, they need to add a review comment requesting the bot. ("@github-actions :ship:")
+#  a) Approve the changes via the 'PR review' feature.
+#     In case you want to approve the PR without updating the version, let the bot know. ("@github-actions :x:")
+#  b) Since PR creators can not approve their own changes, they need to add a 'review comment' requesting the bot. ("@github-actions :ship:")
 #
 # Note: Only repo members can trigger this.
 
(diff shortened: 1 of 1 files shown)


.github/workflows/updateversion.yml
```diff
@@ -14,14 +23,13 @@ jobs:
...
-      ( github.event_name == 'issue_comment' &&
-        github.event.issue.pull_request &&
-        ( github.event.comment.author_association == 'OWNER' ||
-          github.event.comment.author_association == 'MEMBER'
+      ( github.event_name == 'pull_request_review_comment' &&
+        ( github.event.review.author_association == 'OWNER' ||
+          github.event.review.author_association == 'MEMBER'
         ) &&
-        endsWith( github.event.comment.body, '@github-actions update version' )
+        contains( github.event.review.body, '@github-actions :ship:' )
```
tooomm, line 32: @ebbit1q Any reason you used `endsWith` here over the more flexible `contains` as in the other condition?

tooomm (commented):
This is a test review comment

tooomm:
It looks like an code comment does also trigger `  pull_request_review: types: [submitted]`:

Untitled

ebbit1q:
the workflow has to trigger for every comment, there is no other way

tooomm:
>the workflow has to trigger for every comment, there is no other way

Can you explain a bit more, please? I thought this works differently, but am happy to understand better.

I moved the triggers inside the review process and prevented triggers on ordinary comments (like yours above, or mine just now).
At least that was the intention - preventing this empty runs/triggers on every comment created in any PR or issue:
Untitled
.

ebbit1q:
what's the problem with that? the run gets skipped before it even does anything?

tooomm:
Despite the other changes and open questions in this PR, what's the problem with changing/improving upon it?

If you insist on keeping it, sure, I still think some docs are helpful.

ebbit1q:
I notice you've added a link to the wiki in your op https://github.com/Cockatrice/Magic-Token/wiki/Token-updating-process-flow
this guide doesn't seem to work anymore at all, the linked script made by @Psithief doesn't seem to exist anymore, it'd have to be rewritten

PR #146, opened 2022-05-16
Add CC2 zombie token

tooomm:
https://scryfall.com/card/cc2/9/snake-zombie?back
https://api.scryfall.com/cards/13e4832d-8530-4b85-b738-51d0c18f28ec


ebbit1q (approved):
not a bad token I guess

PR #131, opened 2022-02-05
CI: Check image links

tooomm:

PR #148, opened 2022-05-27
Fix all scryfall links

### LINKED ISSUE #92: Scryfall image link changes

tooomm:
Fixes #92

Please double check.


ebbit1q (approved):
seems to work, chance is one of these is somehow broken, you simply did a regex replace and not an api check right? but I'd reckon it's a really small chance and not worth combing through all 489 items here

tooomm:
Yes, that's correct. Just a simple replace.
Did some manual sample tests...

I confirmed that the links are not broken in https://github.com/Cockatrice/Magic-Token/pull/131, but it could be that they somehow link to a wrong image.
But with the same ID that's unlikely.

ebbit1q:
it's happened before, but that was a one in a hundred thousand (that we noticed)

PR #150, opened 2022-06-01
remove misspellings of picURL

ebbit1q:
these were not spelled correctly, causing xmllint to fail and the token for the Dragon Spirit to not have a picture registered.


tooomm (approved):
Good catch!!

(The xmllint run in CI would have spotted it, too! https://github.com/Cockatrice/Magic-Token/actions/runs/2427172644 😉)

PR #153, opened 2022-06-02
Move cmc element for 2/2 shapeshifter

tooomm:


ebbit1q:
I don't see the point of this, the order doesn't matter and both orderings are completely arbitrary

tooomm:
>the order doesn't matter and both orderings are completely arbitrary

Technically it's the same and you are right.

But imagine all values mixed up in the xml file within their scope. Cockatrice wouldn't care... but it would be a mess to work with the file and maintain it.
The tokens involve a lot of manual work by humans still. It helps to keep the quality high if things follow patterns...

ebbit1q:
if you want to propose and document a sorting order for these that we should assume you should do so, I think our focus should rather lie on the amount of other prs that are still drafts right now and get those ready for review.

Issue #85, opened 2020-04-11
Missing tokens


Psithief:
The following tokens were never printed and so do not have official images.
To assist, you will need to provide a URL that allows hot-linking to the image.

1. 0/1 Blue Plant Wall Creature token with Defender named `Kelp`, created by `Wall of Kelp`. (Homelands)
2. \*/\* Black Spirit Creature token created by `Broken Visage`. (Homelands)
3. 1/1 Green Wolf Creature token with "Bands with other creates named Wolves of the Hunt" named `Wolves of the Hunt`, created by `Master of the Hunt`. (Legends)
4. 0/1 Green Wall Creature token with Defender named `Wood`, created by `Jungle Patrol`. (Mirage)
5. 3/3 Green Elemental Creature token created by `Evil Comes to Fruition`. (Archenemy)
6. 4/6 Colorless Golem Artifact creature token created by `The Iron Guardian Stirs`. (Archenemy)

I haven't tested whether cards listed at [mtg.onl](https://mtg.onl/) can be hot-linked to.

Psithief:
Some of the proxies at mtg.onl aren't great image choices or really low image quality, so if other alternatives exist, let me know.

![Kelp](https://www.mtg.onl/static/20e20f54afa5e465046e199265bae73c/4d406/PROXY_Kelp_U_0_1.jpg)

![Spirit](https://www.mtg.onl/static/5681f8f60f717fb528d0d728cab2bd78/4d406/PROXY_Spirit_B_Y_Y.jpg)

![Wolves of the Hunt](https://www.mtg.onl/static/e34edc351ea7ef08c4c4064d1f890731/4d406/PROXY_Wolves_of_the_Hunt_G_1_1.jpg)

![Wood](https://www.mtg.onl/static/70c0c3608291aaee9517eff9cacd43d6/4d406/PROXY_Wood_G_0_1.jpg)

tooomm:
I assume `3/3 Elk` is also one of these:
https://www.reddit.com/r/Cockatrice/comments/g1hvxs/custom_token_33_elk/

Psithief:
This issue is about tokens for old sets that were never printed.

That link is a request for a "token" card for a new set as a convenience. It would need its own issue.

tooomm:
Probably I just have a wrong understanding then.

What's the difference between those old tokens that never were printed, and this new token which was not printed?

[...]

PR #156, opened 2022-06-06
Add art for 3 missing tokens

SlightlyCircuitous:
Add custom art tokens for Broken Visage, Wall of Kelp, and Jungle Patrol. Art is public domain, see comments below for details.

https://github.com/Cockatrice/Magic-Token/issues/85#issuecomment-1146745749

https://github.com/Cockatrice/Magic-Token/issues/85#issuecomment-1146841717


ebbit1q (approved):
thanks!

PR #160, opened 2022-06-16
remove unused token relations

### LINKED ISSUE #161: Funny alternative names
doing a search of these shows we have a couple similar names:
```
"Big Boy Forest Crusher" placeholder for Quartzwood Crasher
"Destoroyah, Perfect Lifeform" alt name for Everquill Phoenix
"What's Kraken" placeholder for Ominous Seas
"Obosh, With The Leggies" placeholder for Obosh, the Preypiercer
"Gigan, Cyberclaw Terror" alt name for Gyruda, Doom of Depths
```

none of these names show up in the xml generated by oracle, they take their primary name and their alt name is ignored, in most cases being overwritten

I propose all of these be removed

_Originally posted by @ebbit1q in https://github.com/Cockatrice/Magic-Token/issues/159#issuecomment-1156826621_

ebbit1q:
fixes #161

I've automated some of the menial tasks in the xml, see the individual commits

most of the changes are purely cosmetic, the only noticeable result to the user is that the phyrexian insect now creates poison counter tokens again


ebbit1q:
because all entries have been sorted they would always be next to each other, so I did not

Psithief:
Wait, why is <related> for the RG Beast preferred over the <reverse-related> on the emblem? I'd have preferred not to repeat the spaces in the Beast token's name more than once.

ebbit1q:
I moved the internal token relations to be <related> instead of <reverse-related> for consistency, this way all <reverse-related> tags refer to the cards.xml

tooomm:
>I moved the internal token relations to be `<related>` instead of `<reverse-related>` for consistency, this way all `<reverse-related>` tags refer to the cards.xml

I saw your description in commit https://github.com/Cockatrice/Magic-Token/pull/160/commits/67b20abc54d557e4449119dde005929a0adce58b regarding that:
```
reverse-related is assumed to only be used for external references to
the xml
related tags should be used for internal links to make this easier and
to avoid certain errors eg the Phyrexian Insect here
```
While we never agreed on a specific way in earlier discussions about related/reverse-related, I don't think this is bad.

Downside is that users can no longer look at the file or a specific token in there and learn about or see all creators at a glance, which I liked. Now, e.g. an emblem creating tokens will not be listed at the resulting token anymore, but the token will be listed at the emblem.

PR #162, opened 2022-06-18
Update Gatherer token links that need special care to Scryfall

SlightlyCircuitous:
I am working on a script to update these all automatically with a find and replace, but the ones with multiple art printings per set trip it up too much so I just did them manually instead of trying to compare pictures algorithmically or something.

Of note, there are other printings of Clue, Food, and Treasure from MH2 in Scryfall but not in the .xml. Dunno if I should add those or not. Also, there is another printing of both Soldier and Copy from CMR, but they are both identical to the ones in the .xml, just with slightly different colors so those are probably not useful.


ebbit1q:
cockatrice doesn't do anything with duplicates of set tags, it just ignores them.
while it's not harmful to have this many SOI  clues, it's fine to just pick one you like.
not to mention many users will have a different set sorted before it and will not use that specific print anyway.

SlightlyCircuitous:
That's what I figured: it just takes the first image listed for each set or something. No need for me to add the other MH2 token art then, unless someone really likes them. Just thought I would ask. I will stick to just replacing every Gatherer link with a Scryfall link as asked for in the issue and leave the question of picking a favorite for the multiple prints to someone else. 
Unless you are implying that you don't want to merge this and I should just do all of them in one go with a script and solve the multiple art issue by letting the script pick one for a different pull request.

ebbit1q:
if someone really likes a specific art their option is to use the custom image folder.

it is the privilege of the person submitting the pull request to pick the art used, they are the one that made the pr after all, we're not going to go back and change them just to pick a different art

SlightlyCircuitous:
Ah, sorry, I meant that since all of the duplicate art is in the file already, I am trying to make the fewest changes possible by just replacing the links and not removing any lines. By 'leave it to someone else', I meant leave it to some other maintainer to make the decision on whether or not to get rid of duplicates, not constantly edit the art through PR's. But clearly I am being overly concerned with making too many edits and you've been pretty clear on what you think should be done with duplicates anyway.

I'll pick my favorites here (easier to edit this PR where I can see all the art rather than deal with duplicates algorithmically) and get rid of the rest.

tooomm:
Hmm. As I understood it, the project and @Psithief always aimed for this to be a comprehensive database of all printed tokens - including different art variants within a set.
We added duplicates on purpose, only their order was basically up to the PR creator.

SlightlyCircuitous:
If that is true, I am happy to edit my PR again to un-delete the duplicates (or you can just merge the original commit? I'm not exactly sure how merging works). I simply want to update the links (and move that one zombie token) and will abide by any consensus.

SlightlyCircuitous:
Reversed my previous commit of removing duplicates and took the opportunity to update the links of other tokens that are hard to deal with (because they have card types or card names on Scryfall, mostly). This doesn't cover emblems, because I think I can do those automatically with a little more work.

Psithief:
> Hmm. As I understood it, the project and @Psithief always aimed for this to be a comprehensive database of all printed tokens - including different art variants within a set. Cockatrice could also use the data at one point. We added duplicates on purpose, only their order was basically up to the PR creator.

I'll explain, I wanted to give people all of the URLs so they could find the URL of their favourite token from the tokens file, download the file and put it in their custom card image directory.

ebbit1q:
I don't think many people would do that, it's a lot easier to simply look up the token on scryfall than it is to search through the xml

tooomm:
Thanks for chiming in psithief!

I see no urge to remove the existing data. Unless we want/have to trim the file for some reason, or it solves some issue?

ebbit1q (approved):
I guess we are keeping the duplicates for now

Psithief:
> I don't think many people would do that, it's a lot easier to simply look up the token on scryfall than it is to search through the xml

It was before Scryfall existed, @ebbit1q 😁

ebbit1q:
yes, and scryfall exists now

Issue #149, opened 2022-05-31
Update image links from wizards to scryfall


tooomm:
Picture quality and resolution is generally (way) worse on Wizards.

First example from the tokens file: `Ajani's Pridemade (Token)`
[Wizards](https://media.wizards.com/2019/m20/en_CG2GSQIUiW.png) vs. [Scryfall](https://c1.scryfall.com/file/scryfall-cards/large/front/b/0/b0819e8e-fb7e-43c7-a7cf-d768f43193ac.jpg?1592515934)

Gatherer used to be down sometimes as well.

PR #163, opened 2022-07-01
De-Gather all Remaining Links

SlightlyCircuitous:
Now that all the complicated cards were taken care of in https://github.com/Cockatrice/Magic-Token/pull/162, I have run my two algorithms to replace all the other links automatically. I have attached a .txt file (github wouldn't accept a .py) containing the algorithms (and tests) I used to do this.

[degathering_functions.txt](https://github.com/Cockatrice/Magic-Token/files/9031948/degathering_functions.txt)

I borrowed some code from Magic-Spoiler to call the Scryfall API and so, if I am reading the GPL correctly, I had to license that function under GPL; the rest should be able to be (and are) public domain. Please correct me if I have misapplied the license or should not have used this code.

This should close https://github.com/Cockatrice/Magic-Token/issues/149.


ebbit1q:
the results of running gpl code are not covered by the license, only the code and binary representations of it are, there is no problem including such automated changes in the xml

ebbit1q (approved):

Psithief:
Useless fact of the day:
The media.wizards.com links are not Gatherer links. Also, tokens didn't get added to Gatherer, except in a number of Duel Decks they added by mistake. 😀

ebbit1q:
true, they are actually from their blog

SlightlyCircuitous:
Damn, that really ruins the "Magic: De-gathering" pun I was using for all my file names. Oh well, they're gone now, whatever they are.

PR #164, opened 2022-07-03
Adds NCC Tokens

mainman879:
When https://github.com/Cockatrice/Magic-Token/pull/141 added the SNC set, it did not include the corresponding Commander Set (NCC). This PR adds these tokens relations.


tokens.xml
```diff
@@ -1387,16 +1387,25 @@ This creature can block only creatures with flying.</text>
...
             <reverse-related attach="attach">Citizen's Crowbar</reverse-related>
             <reverse-related>Courier's Briefcase</reverse-related>
             <reverse-related>Darling of the Masses</reverse-related>
             <reverse-related>Exhibition Magician</reverse-related>
+            <reverse-related count="x">Grand Crescendo</reverse-related>
             <reverse-related>Halo Fountain</reverse-related>
+            <reverse-related count="x">Indulge // Excess</reverse-related>
+            <reverse-related>Kitt Kanto, Mayhem Diva</reverse-related>
             <reverse-related>Luxurious Libation</reverse-related>
+            <reverse-related count="x">Master of Ceremonies</reverse-related>
```
ebbit1q, line 1402: should this be an x? the other player will just make one token

tokens.xml
```diff
@@ -8533,13 +8564,15 @@ This creature can't be enchanted.</text>
             <reverse-related>Tempting Contract</reverse-related>
             <reverse-related>Thieves' Tools</reverse-related>
             <reverse-related>Tireless Provisioner</reverse-related>
+            <reverse-related count="x">Tivit, Seller of Secrets</reverse-related>
```
ebbit1q, line 8567: interestingly, because neither of these is excluded and both require x, the client does nothing at all when you try to use the create all tokens action

tokens.xml
mainman879: The player that controls the Master of Ceremonies could create multiple tokens, as they create 1 for each opponent that chooses that specific mode. Could maybe make it 1 by default and count=x as an excluded.

ebbit1q: I think that would be more intuitive yeah, for fast gameplay your opponents can simply make their own tokens and you copy them

ebbit1q:
```suggestion
            <reverse-related>Master of Ceremonies</reverse-related>
            <reverse-related count="x" exclude="exclude">Master of Ceremonies</reverse-related>
```

tokens.xml
mainman879: Well that is awkward (and should maybe be raised as an issue on the cockatrice github). I suppose just making one or the other exclude is best.

ebbit1q (approved):

tokens.xml
Psithief: There has never been a card like this one before! Ouch.

Issue #48, opened 2017-11-30
Upvote for silver-border sets.


Psithief:
I'm not interested in doing tokens for silver-bordered sets. 

Do a thumbs-up reaction to this post to show demand for this.

tooomm:
Did the former UN-sets have tokens, too?

Do you add the other (normal) tokens `UST` includes?
They have damn sweet full art token for angel, spirit, zombie and other common stuff.
https://magic.wizards.com/en/articles/archive/news/tokens-unstable-2017-11-21

Thunderbeliever:
Thumbs down

Psithief:
> Did the former UN-sets have tokens, too?

Former `Un-` sets went without errata for so long that they were too hard to work with.

> Do you add the other (normal) tokens UST includes?

Yes, but perhaps as part of the Rivals of Ixalan update.

PR #165, opened 2022-07-04
sort sets and cards

ebbit1q:
more bikeshedding, similar to #160 , check individual commits for more info

PR #166, opened 2022-07-04
Add Un-Set Tokens

SlightlyCircuitous:
Closes https://github.com/Cockatrice/Magic-Token/issues/48

It seemed like these should be in here since all the UN-cards are in cards.xml and they need their token buddies. Given that there was a fair bit of Un-Set errata when Unstable came out, everything was pretty easy to deal with.

I am also happy to commit to updating the xml when Unfinity comes out later this year.


tokens.xml
```diff
@@ -695,6 +695,20 @@ Creature tokens you control have flying and vigilance.</text>
...
+        <card>
+            <name>Beeble Token</name>
+            <prop>
+                <colors>U</colors>
+                <type>Token Creature — Beeble</type>
+                <maintype>Creature</maintype>
+                <pt>1/1</pt>
+            </prop>
+            <set picURL="https://c1.scryfall.com/file/scryfall-cards/large/front/9/b/9bbfa4d4-106b-4fec-86ac-b81b71a8f432.jpg?1584741596">UND</set>
+            <reverse-related>B.O.B. (Bevy of Beebles)</reverse-related>
```
ebbit1q, line 707:
```suggestion
```
bob makes 4 beebles, it doesn't have an option to make only one

tokens.xml
```diff
@@ -3091,6 +3135,21 @@ This creature's power and toughness are each equal to the number of instant and
             <token>1</token>
             <tablerow>2</tablerow>
         </card>
+        <card>
+            <name>Faerie Spy Token</name>
+            <text>Flying, haste
+                Whenever this creature deals combat damage to a player, draw a card.</text>
```
ebbit1q, line 3141:
```suggestion
Whenever this creature deals combat damage to a player, draw a card.</text>
```
card text should not be indented, it gets copied over:
image

tokens.xml
```diff
@@ -8114,6 +8222,20 @@ Equip {0}</text>
             <token>1</token>
             <tablerow>2</tablerow>
         </card>
+        <card>
+            <name>Storm Crow Token</name>
```
ebbit1q, line 8226:
```suggestion
            <name>Storm Crow </name>
```
the card specifies that the card is named "Storm Crow" this means it does not get Token added to the name officially, yet in order to distinguish it from Storm Crow the card we need to add a space

tokens.xml
```diff
@@ -2070,6 +2098,22 @@ At the beginning of your upkeep, sacrifice another creature. If you can't, this
             <token>1</token>
             <tablerow>2</tablerow>
         </card>
+        <card>
+            <name>Dragon Token     </name>
+            <text>Flying</text>
+            <prop>
+                <colors></colors> <!--This card's color is Gold-->
```
ebbit1q, line 2105: while I agree that adding it to the real color property is undesirable this comment feels out of place, either add it to the card text (might be actually useful) or just omit it

tokens.xml
SlightlyCircuitous: The option to make only one is to satisfy the unique way B.O.B. handles loyalty counters:
"The number of loyalty counters on B.O.B. is equal to the number of Beebles you control. (Create or sacrifice Beebles whenever B.O.B. gains or loses loyalty.)"
Since you can add one loyalty with it's first loyalty ability, you need to be able to create one Beeble at a time.

tokens.xml
SlightlyCircuitous: Agreed. Made it a parenthetical in the text field.

tokens.xml
ebbit1q: that's a good point, I didn't interpret the card fully

tokens.xml
ebbit1q: I like the new addition, technically tokens aren't cards but I think this is fine

tokens.xml
```diff
@@ -11429,6 +11554,18 @@ Whenever a creature deals combat damage to you, its controller becomes the monar
             <token>1</token>
             <tablerow>1</tablerow>
         </card>
+        <card>
+            <name>Acorn Stash</name>
```
ebbit1q, line 11558: I think because of the similarity to experience counters I think it should be put next to those, with the type Counter

ebbit1q: I'm also not sure on the way to order these "extra" tokens, if you have ideas on unnecessarily rigidifying the structure of the tokens.xml leave a comment on my pr #165

SlightlyCircuitous: I moved it to be alphabetical with the Counters and adjusted it's type and maintype accordingly.

ebbit1q (approved):

ebbit1q:
not that the exclude on the counters isn't actually necessary as the card makes no other tokens (I think that got in as a bit of lazyness in the past, it doesn't actually hurt but doesn't exclude the entry counterintuitively)

Psithief:
> note that the exclude on the counters isn't actually necessary as the card makes no other tokens (I think that got in as a bit of lazyness in the past, it doesn't really hurt but doesn't actually exclude the entry counterintuitively)

I coded it that way as a 'garbage in, sensible default out' scenario.

PR #167, opened 2022-07-06
Fix Undercity image

tooomm:
Spotted trough improved duplicate link listing via https://github.com/Cockatrice/Magic-Token/pull/131/commits/42440b7921d71da279ff01bed741d0430637e6c5


ebbit1q (approved):
good catch

PR #168, opened 2022-07-07
imgur --> scryfall

tooomm:
Looks like most images hosted at imgur.com are from older sets, that scryfall has no token for?

Found only those two.

Same is true for staticneo.com hosted images. Found none there.


ebbit1q:
we have some tokens that are fanmade yeah

ebbit1q (approved):
small increase in quality

SlightlyCircuitous:
Wizards printed the first tokens in Unglued (1998) so there are two dozen or so old sets (between '93 and '98) without printed tokens. Though, reprint sets like Ultimate Masters and Double Masters are starting to fill in the gaps. I plan to do a PR with a few tokens I found in those to replace some fanmade ones as soon as Scryfall puts up high res images for Double Masters.

ebbit1q:
scryfall's scans of 2xm tokens look just as high res as their other cards to me, which of these tokens hasn't been printed before though? I don't see any unique ones

SlightlyCircuitous:
The scryfall jsons for each Double Masters token says 'lowres' and 'highres: false' so I assume better ones are coming, though they do look fine enough now, it's true. Also, Double Master has a Phyrexian Golem token that's actually called 'Phyrexian Golem' so I wanted to swap that one in since it has official errata.

ebbit1q:
ah, quadruple masters, yeah, those are lowres, each of these has been printed before though

if you want to add 2x2 cards please add all tokens at the same time

SlightlyCircuitous:
ok, I'll just do the other ones first then and do 2x2 later

PR #170, opened 2022-07-07
Add missing MH2 images

tooomm:
Credit: https://github.com/Cockatrice/Magic-Token/pull/162#issue-1275642427


ebbit1q:
why add these? we already had MH2 art

mainman879:
Having multiple image links under the same set does effectively nothing, as it will just use the first one in the list to my knowledge. Seems "fine" for completeness sake and won't hurt anything, but doesn't really do anything either.

tooomm:
> why add these? we already had MH2 art

Those are obviously different, missing ones.

The PR follows the current convention on listing all available art.
It's not the first discussion about the topic. As long as there is no common decision on changing this, I see no issue in completing the list once somebody stumbles upon missing entries.

If there is a decision on only listing one image per set (let the creator pick), we should clean up the file and be ok with Cockatrice never using such data, e.g. allow users to pick token art on creation in a game (`Create Token` dialog) or in deck editor.

<br>

You are correct, mainman. My PR completes the list of available art, but does not affect Cockatrice currently.

PR #171, opened 2022-07-09
Add VOC and UMA Tokens

SlightlyCircuitous:
Same idea as https://github.com/Cockatrice/Magic-Token/pull/168

Main idea is to replace several of the fanmade tokens with official printings; figured I should add the whole set if I was going to add a few though.


ebbit1q (approved):
seems good, thanks for moving some of the tokens as well

PR #172, opened 2022-07-12
Add RNA Guildkit (GK1+GK2) Tokens

SlightlyCircuitous:
Continuing on https://github.com/Cockatrice/Magic-Token/pull/171, replaces a few more fanmade tokens.


SlightlyCircuitous:
I did not add the 2/1 R Goblin token from GK2 because it honestly seems like a misprint. It is, in theory, the official goblin token for Rakdos Guildmage, but it doesn't have Haste like it is supposed to. According to Scryfall, a 2/1 R Goblin without haste isn't useful for any other card and the xml didn't have a line for one without haste already so I didn't bother adding it.

ebbit1q:
> I did not add the 2/1 R Goblin token from GK2 because it honestly seems like a misprint. It is, in theory, the official goblin token for Rakdos Guildmage, but it doesn't have Haste like it is supposed to. According to Scryfall, a 2/1 R Goblin without haste isn't useful for any other card and the xml didn't have a line for one without haste already so I didn't bother adding it.

wizards has been very inconsistent on whether or not haste is printed on the tokens, the first token with haste has it printed on the token, after that they made a policy to not print haste on the tokens, then with the introduction of cards in eldraine it became relevant again so they reverted this policy, this means most tokens with haste that have been printed to date don't actually have haste printed on the token.

SlightlyCircuitous:
Ah, I see. That's extremely strange. I will add that Goblin token in then.

tooomm (approved):
Looking good to me! 👍 

Thanks for also removing other user tokens that have official tokens in between, and fixing some other wrong references along the way.

ebbit1q:
note, things like https://cdn.staticneo.com/w/mtg/2/2d/Cat_Warrior.jpg are official tokens, not fanmade, they aren't on scryfall for some reason but these tokens often come from wotc blog posts

SlightlyCircuitous:
Ah, I see. That explains why some of them have official art then.

PR #174, opened 2022-07-27
Fix ISD Ooze token picture

tooomm:
Image started to fail recently.

Proudly sponsored by https://github.com/Cockatrice/Magic-Token/actions/runs/2745469559#summary-7537058181


tooomm:
Moving on with this fix.

@github-actions update version

PR #175, opened 2022-08-10
Rename challenge file

tooomm:
Follow up to #173.


ebbit1q:
calling them decks is confusing, the file doesn't contain the decks, which you would have to get separately, it just contains the card objects

tooomm:
True! It contains the cards the decks consist of. Do you have another good suggestion?

"challenge_cards"?

ebbit1q:
I don't think it's worth renaming it at all, the small increase in clarity would mean existing links to the file would stop working

tooomm:
The file was broken for over 5 years, I don't think we have to care about that too much. 😃

If others think this is not useful, we can close this as well.

PR #176, opened 2022-08-15
Adds Foretell maintype of State

mainman879:
This just adds a line to add maintype of State so it matches The Initiative, On an Adventure, etc.


ebbit1q (approved):
ok

PR #179, opened 2022-09-26
Add Tokens from DMU

SlightlyCircuitous:
Fixes https://github.com/Cockatrice/Magic-Token/issues/178


ebbit1q (approved):
we really needed these, thanks!

PR #180, opened 2022-09-26
Create Token XML Updater

SlightlyCircuitous:
This is the script that I mentioned in the discord for automating some parts of adding new tokens to the XML. Hopefully other may find it useful. I plan to update it to run in the cmd line soon, just figured it would be better to post the code sooner rather than later.


SlightlyCircuitous:
Closed since it actually needs lots more work

PR #181, opened 2022-09-26
Fix typos in DMU tokens

SlightlyCircuitous:
Noticed these while doing DMC


ebbit1q (approved):
I should have paid closer attention when loading the xml when reviewing

PR #184, opened 2022-09-28
Create Token XML Updater

SlightlyCircuitous:
Second try opening a PR for this. Think I smoothed out all the formatting issues, and it runs in the cmd line now.


Psithief:
I think this should actually be in a separate repository.
This is actual executable (interpreted) code, but I created this repository was created under the `Unlicense` because it didn't have any code.
 @ebbit1q @tooomm ?

Psithief:
Perhaps a licence like GPL-3.0 like https://github.com/Cockatrice/Magic-Spoiler would be appropriate ?

Psithief:
@ZeldaZach I think this needs to be in a separate repository with a proper licence.

SlightlyCircuitous:
Whatever y'all feel is appropriate is fine with me. GPL-3.0 seems like a good choice as part of the code is already licensed under that anyway since I borrowed from Magic Spoiler for the Scryfall API bits.

SlightlyCircuitous:
I accidentally closed this while messing around with the branch. But I'm not sure where it stands anyway, so maybe its for the best.

ebbit1q:
I'd say a separate repo would be more focused, I think we have enough monolithic repos as is :^)

Psithief:
I think Cockatrice would have to fork a new repository with just this script from you, @SlightlyCircuitous.

Psithief:
@ZeldaZach is that the best way of doing things?

tooomm:
What about this?
- Rename the current repo to `Magic-Token_xml`, `Magic-Token_old` or similar
- Archive the repo
- Create a new repo with same name again (`Magic-Token`), add GPL-3.0, add team again
- SlightlyCircuitous creates a PR with the script

Or do you really want to separate stuff completely and have two repos? Even if it's just a single script?
I think it works quite good in the spoiler repo as is?

SlightlyCircuitous:
In case its relevant, the current version of this script lives [here](https://github.com/SlightlyCircuitous/update-token-xml) now. If anyone wants to make a new repo or whatever y'all decide, feel free to just copy it wholesale from there.

PR #188, opened 2022-11-11
Remove Parentheses Around Emblem Names

SlightlyCircuitous:


Psithief:
Why is there no description for this? Why do you want it changed?

ebbit1q:
ah, we discussed this on discord, it looks more uniform with all of the tokens that end with Token and it's the same way they are listed on scryfall.

SlightlyCircuitous:
yes, what ebbit said. It's about uniformity with Scryfall, mainly. I apologize for not including a description in the first place.

ebbit1q:
no news is good news, merging this

PR #189, opened 2022-11-29
CI: Bump version for image check

tooomm:
The used action and lychee tool fixed several issues in the last few releases.
This now ensures, that we stay up to date with fixes etc. while not running into breaking changes.


ebbit1q:
cache working quite well: https://github.com/Cockatrice/Magic-Token/actions/runs/3579782185/jobs/6021307977

PR #191, opened 2022-12-03
change occult epiphany to use x

ebbit1q:
change occult epiphany to use x


tooomm (approved):
Jep, that sounds about right. 👍

PR #192, opened 2022-12-04
CI: Show useful data about token URLs

tooomm:
We automatically check if imagine links are accessible/available.

This adds some more insights and helps with understanding the file. It gives us a little overview of its content and our sources/dependencies, also highlighting doubled links (which we eliminated by now) and completly missing card art.

Result:
https://github.com/Cockatrice/Magic-Token/actions/runs/14453291487?pr=192#summary-40530736489

Screenshot:
| March 2024 | August 2024 | November 2024
| - | - | - |
|  |  |


tooomm:
The upstream issue is resolved and `dump` works as expected now.

.github/workflows/check_links.yml
```diff
@@ -31,14 +31,42 @@ jobs:
           key: cache-lychee-${{ github.sha }}
           restore-keys: cache-lychee-
 
-      # Extract and check URLs directly from token files
+      # Extract and check URLs directly from token files (checking dumped file did not work)
       - name: Check token art URLs
         uses: lycheeverse/lychee-action@v1
         with:
-          args: '${{env.lychee_args}} -- tokens.xml challenge_tokens.xml'
```
ebbit1q, line 38: is this intentional?

.github/workflows/check_links.yml
```diff
@@ -31,14 +31,42 @@ jobs:
...
+          grep . lychee/out.md | sed 's/\.jpg?.*/.jpg/' | sort | uniq -cd >> $GITHUB_STEP_SUMMARY
+
+      # Analyse extracted links (2/2)
+      - name: List image hosting sources
+        if: steps.tokens_pic_urls.outcome == 'success'
+        shell: bash
+        # Extract domains from URLs | remove blank lines | sort | count and list | sort descending
+        run: |
+          echo "📊 **Image Hosting Statistics**" >> $GITHUB_STEP_SUMMARY
+          awk -F/ '{print $3}' lychee/out.md | grep . | sort | uniq -c | sort -nr >> $GITHUB_STEP_SUMMARY
```
ebbit1q, line 68: why is grep . here? is there empty lines? why does it run sort twice?

.github/workflows/check_links.yml
tooomm: Yes, we no longer have a separate challenge_tokens.xml file.
See #225

.github/workflows/check_links.yml
```diff
@@ -31,14 +31,42 @@ jobs:
...
+          args: '${{env.lychee_args}} -- tokens.xml'
           fail: true
           jobSummary: true
 
+      # Extract picture URLs from tokens.xml
+      - name: Extract URLs
+        id: tokens_pic_urls
+        uses: lycheeverse/lychee-action@v1
+        with:
+          args: '--dump --exclude www.w3.org -- tokens.xml'
```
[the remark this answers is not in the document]
tooomm: Looked into the docs again. As I understand it, link checking is the default behavior and providing `dump` will perform the extraction of the links and writes them, but skips checking.

Both at the same time is currently not possible.
Also too bad I could not make the link check from the dumped file work and reuse that.
Maybe I reinvestigate at a later point again.

.github/workflows/check_links.yml
tooomm: >why is grep . here? is there empty lines? 

There used to be empty lines in the output when we had token or sets added without image links. This is currently not the case.
Since we do not validate for empty/missing links that could happen again.

But I can also remove it.
Am actually not sure if empty lines would even mess with the result or not.

<br>

>why does it run sort twice?

You mean sorting descending directly before count + list?
I did simply not think about combining that.

.github/workflows/check_links.yml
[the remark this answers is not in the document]
ebbit1q: ok

.github/workflows/check_links.yml
ebbit1q: I guess the comment on it could be more explicit, it takes the third column from the file, removes empty lines, sorts and then counts the duplicates, listing them with their count before the url, then sorts again but based on the number that represents the amount of duplicates descending.

tooomm: Updated the wording of the comment.

The sorting needs to be split because first it's sorting to bring same domain names next to reach other for counting to work. Only after combining the entries and adding quantity numbers it can be sorted descending.

tooomm:
Issue resolved with upstream

PR #194, opened 2023-01-28
Correct Mordenkainen Emblem

SlightlyCircuitous:
The Mordenkainen emblem had the Mu Yanling emblem text for some reason. Fixed that.


ebbit1q (approved):
thanks for noticing that 👍

PR #195, opened 2023-01-30
Add DMR Tokens

SlightlyCircuitous:
Adds DMR tokens, mostly to replace the 2 staticneo links with scryfall links.


tokens.xml
```diff
@@ -1334,7 +1335,7 @@ This creature can't block.</text>
                 <cmc>0</cmc>
                 <pt>2/1</pt>
             </prop>
-            <set picURL="https://cdn.staticneo.com/w/mtg/0/08/Cat2.jpg">APC</set>
+            <set picURL="https://cards.scryfall.io/large/front/0/5/05a61555-ddb9-4d55-b637-a219d772033d.jpg?1674397379">DMR</set>
```
ebbit1q, line 1338: I'm still not certain why we are replacing these, they work and are high quality pics that scryfall doesn't have, I'm pretty certain they are crops from mtgo, where these tokens got released.

I guess the mtg.neoseeker.com wiki isn't the most credible source nor do we have any guarantee of its continued existance.

though for some reason it seems like all token scans on scryfall since SNC have been crappy quality.

SlightlyCircuitous: It just seemed to me like we were very much favoring Scryfall over other sources based on past PRs I've seen merged, but if you'd like to keep the neoseeker scans, that's fine with me too. Easy enough to leave those lines in as an option.

ebbit1q: well, I'd really just like scryfall to do better scans for these tokens, they look all fuzzy

SlightlyCircuitous: I've put them back in since I agree with you that they are good images of those tokens.

SlightlyCircuitous: Although, since they are all reprints, we might not even need to put these in at all. They don't add much beyond replacing those links.

ebbit1q: they're also cropped differently etc. they fit in less well, but they aren't blurry which I guess just comes down to personal preference

ebbit1q (approved):

PR #196, opened 2023-01-31
Add ONE Tokens

SlightlyCircuitous:
* Adds 10 new tokens and 3 art lines for reprints
* Adds missing set line to Sketch token
* Includes reverse relations to ONE and ONC cards (but no ONC token entries)
* Removes Goblin art from Phyrexian Goblin entry as we now have a printed Phyrexian Goblin token and that goblin art already appears in a Goblin Token entry in the file
* Did not remove any Golem art from the Phyrexian Golem entry as that art does not appear anywhere else in the file and that token was reprinted in 2XM as 'Phyreixan Golem'
* Did not do reverse relations for the Poison counter for the 57 new cards with toxic, but I will if you ask me to


SlightlyCircuitous:
Thank you for adding those relations for me.

PR #200, opened 2023-03-07
Add Karn, Living Legacy emblem from Arena

SlightlyCircuitous:
Adds missing art for a token mentioned in issue https://github.com/Cockatrice/Magic-Token/issues/85#issuecomment-1269095070

I took screenshots at the highest graphics Arena would let me have and then did my best with the crop. Please let me know if I need to alter either image further.


SlightlyCircuitous:
I just noticed that I have some of these images uploaded as jpegs and some as pngs; is there a perferred format?

ebbit1q:
in my opinion using lossy formats like jpeg is fine, you don't really notice the difference in quality as long as the image is large enough and the size difference is quite substantial even if you were to use jpegs with more pixels.

ebbit1q:
the jpeg version looks off, why are the corners purple? I think the png was better

SlightlyCircuitous:
I am not sure why the jpeg corners are purple, you will have to ask GIMP. I agree with you that the png looks better anyway (it's brighter) so I have changed it back.

tokens.xml
```diff
@@ -11593,7 +11593,7 @@ Throne of the Dead Three — Reveal the top ten cards of your library. Put a cre
         <card>
             <name>Kaito Shizuki Emblem</name>
             <text>Whenever a creature you control deals combat damage to a player, search your library for a blue or black creature card, put it onto the battlefield, then shuffle.</text>
-            <prop>
+            <prop>https://github.com/SlightlyCircuitous/image-storage/blob/main/karn_emblem_arena.png?raw=true
```
ebbit1q, line 11596: what's this?

ebbit1q:
```suggestion
            <prop>
```

SlightlyCircuitous: Not sure. must have accidentally added it pasted it there when I was adding the png link back.

ebbit1q (approved):
cool beans

Issue #201, opened 2023-03-07
Move to *.scryfall.io for card sources


Elarnon:
Scryfall changed its image URIs: https://scryfall.com/blog/upcoming-api-changes-to-scryfall-image-uris-and-download-uris-224

Magic-Token still uses old c1.scryfall.com URIs instead of the new cards.scryfall.io URIs.

SlightlyCircuitous:
Is this as simple as changing all instances of "c1.scryfall.com/file/scryfall-cards" to "cards.scryfall.io" and leaving the rest of the url the same?

Elarnon:
I believe that should work, at least as a quick fix, although ideally it'd be better to use an api.scryfall.com url (possibly with an accompanying script to resolve the redirects) in case the urls change again.

(More specifically, cards.scryfall.io URLs are aggressively cached and api.scryfall.com URLs do appropriate cache busting)

SlightlyCircuitous:
yeah, I'm not super familiar with how the code that reads the token.xml works atm so I don't know what (if anything) would need to change to use the API instead.

ebbit1q:
to have the token.xml use the api all that needs to change is the picurl in the xml, cockatrice will resolve redirects as normal

tooomm:
> cockatrice will resolve redirects as normal

I'm wondering if the API rate limits could be an issue for Cockatrice as well?

Like if somebody loads a new deck and all card pictures are queried...
>**Rate Limits and Good Citizenship**
>
>We kindly ask that you insert 50 – 100 milliseconds of delay between the requests you send to the server at api.scryfall.com. (i.e., 10 requests per second on average).
>
>Submitting excessive requests to the server may result in a HTTP 429 Too Many Requests status code. Continuing to overload the API after this point may result in a temporary or permanent ban of your IP address.
>
>The file origins used by the API, located at *.scryfall.io do not have these rate limits.

ebbit1q:
they are, it's just rather hard to diagnose, also cockatrice's picture loader is pretty slow and will never do requests in parallel. I often ask for logs when I suspect this has occurred to a user in the support channel on discord yet never see a 429 code, truncating often gets in the way here though so I don't actually have that many instances where I can check this. not to mention cockatrice will fall back to using gatherer when such an error would occur.

[...]

PR #204, opened 2023-03-10
Update Token Links to Use API calls

SlightlyCircuitous:
Fixes https://github.com/Cockatrice/Magic-Token/issues/201 and should future proof the file against further link change. 

API requests adheres to Scryfall's [/cards/:id endpoint format](https://scryfall.com/docs/api/cards/id). Image size defaults to large and face defaults to front so only format=image and the occasional face=back need to be specified.

I have checked that this file works in Cockatrice (at least for as many tokens as I could stand to look at) but its a huge number of changes so its not impossible that something somewhere broke. I also specifically checked all the back sides had well-formed requests since they required extra bits stuck on the end.

### DIFF
--- a/tokens.xml
+++ b/tokens.xml
@@ -40,7 +40,7 @@
                 <cmc>0</cmc>
                 <pt>4/4</pt>
             </prop>
-            <set picURL="https://c1.scryfall.com/file/scryfall-cards/large/front/7/1/71dc8556-a658-40e1-8a93-6a62af208a28.jpg">HOU</set>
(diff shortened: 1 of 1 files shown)


SlightlyCircuitous:
Should I close this in favor of https://github.com/Cockatrice/Magic-Token/pull/209?

ebbit1q:
I think so, it seems like the easier solution

SlightlyCircuitous:
OK, will do.

PR #206, opened 2023-03-11
set xsi to http instead of https

### LINKED ISSUE #202: Tokens.xml fails to validate
When running `xmllint --schema ./cards.xsd ./tokens.xml` using the current v4 cards.xsd, I receive the following error:

`./tokens.xml:7: element cockatrice_carddatabase: Schemas validity error : Element 'cockatrice_carddatabase', attribute '{https://www.w3.org/2001/XMLSchema-instance}noNamespaceSchemaLocation': The attribute '{https://www.w3.org/2001/XMLSchema-instance}noNamespaceSchemaLocation' is not allowed.`

Am i doing something wrong or is this an issue of expectations mismatch between cards.xsd and tokens.xml?

I have been simply ignoring this message because I know the xml works fine with that attribute, but it would be nice if the validation tool would tell me everything is OK when I run it with a valid file.

ebbit1q:
fixes #202


tooomm (approved):
Nice way with the lycheeignore file. 👍 

Alternative would be to add `--exclude www.w3.org` or similar to the args.

ebbit1q:
this was the way to exclude urls listed in the action's readme, if you prefer the arg that'd also be fine, it'd keep the workflow confined to its own file instead of cluttering the repo.

tooomm:
Probably there won't be any other links added there, right?
In that case --exclude seems cleaner.

tooomm (changes requested):
I guess the `continue-on-error` behaves different than you think and it's not helpful here:
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idcontinue-on-error

We only have a single job that is run (check_urls)...

But just saw that you wanted to get rid of the `if: always()` that way.
Even putting it [on the step level](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error) would only make the workflow pass, but still skip following steps I think. Meaning if the check on tokens.xml fails, challenge_tokens.xml is never checked. It's probably ok, to only put it on the second check - just for consistency I put it to both.
Or did you test this?

ebbit1q:
hmm, I might have misinterpreted the docs then, why don't we have one step checking both files? this would marginally improve speed too.

ebbit1q:
:+1:

tooomm (approved):
Good idea to combine the two checks in one run!

Now the env is less useful, but it still makes it a bit more readable I think.

PR #207, opened 2023-03-11
Add J22 Tokens

SlightlyCircuitous:
*Adds artless entries for 3 new tokens (no physical print available)

*Adds reverse-related lines to 5 existing tokens

*Adds excluded count="x" for Zendikar's Roil to allow for making multiple tokens at once off multiple lands triggering landfall 

*Replaces "x=1" with "x" for Iridescent Hornbeetle to conform with style rules


ebbit1q (approved):
thanks

PR #209, opened 2023-03-12
Add script to update picture links for tokens from Scryfall API

### LINKED ISSUE #201: Move to *.scryfall.io for card sources

Elarnon:
Fixes #201

This script changes the links from the old CND on c1.scryfall.com to links to the new CDN on cards.scryfall.io.

The refresh.py script can read a tokens.xml file containing CDN picURLs of either version, and makes requests to the Scryfall API to generate the most up-to-date URLs for each token.

The script can be run with:

  $ python3 refresh.py -i tokens.xml
  $ python3 refresh.py -i challenge_tokens.xml

This avoids using api.scryfall.com URLs for the tokens (which are rate-limited) while still being able to update the file with new URLs by running the script.

### DIFF
--- a/refresh.py
+++ b/refresh.py
@@ -1,0 +1,185 @@
+"""
+This script parses a token.xml file, collects the picURLs within, and replaces
+the URLs to Scryfall cards with up-to-date URLs by querying Scryfall's API.
+"""
+
+from xml.sax import saxutils, make_parser, handler
+from urllib.parse import urlsplit
+from urllib.request import Request, urlopen
+import itertools
+import json
+import sys
+import time
+import os
+import tempfile
+import pathlib
+import shutil
+
+SCRYFALL_MAX_LIST_SIZE = 75
+
+def cards_collection(identifiers):
+    """
+    Get information about a set of cards using the Scryfall API.
+
+    This simply returns a list of dictionaries representing the Card objects as
+    returned by the /card/collection Scryfall API.
+
+    If the list of identifiers is larger than Scryfall's API limit,
+    cards_collection automatically splits the list into smaller chunks and
+    makes multiple requests.
+    """
+
+    start_time = 0
+    n = 0
+    while n < len(identifiers):
+        chunk = identifiers[n:n + SCRYFALL_MAX_LIST_SIZE]
+        print("Requesting chunk {}-{}/{}...".format(n, n + len(chunk), len(identifiers)))
+        n += SCRYFALL_MAX_LIST_SIZE
+
+        payload = json.dumps({'identifiers': chunk}).encode('utf-8')
+        req = Request('https://api.scryfall.com/cards/collection', payload,
+                      headers={'Content-Type': 'application/json'})
+
+        # Rate limiting
+        cur_time = time.time()
+        delta_time = cur_time - start_time
+        if delta_time < 0.1:
+            time.sleep(0.1 - delta_time)
+        start_time = time.time()
+
+        with urlopen(req) as f:
+            list_obj = json.load(f)
+            assert not list_obj.get('has_more', False)
+            assert 'warnings' not in list_obj
+
+            yield from list_obj['data']
+
+def parse_picurl(picurl):
+    """
+    Parse a Scryfall picURL into its components.
+
+    The Scryfall picURL must be in one of those forms:
+
+        - https://c1.scryfall.com/file/scryfall-cards/<version>/<face>/*/*/<uuid>.jpg
+        - https://cards.scryfall.io/<version>/<face>/*/*/<uuid>.jpg
+
+    If it is, a dictionary with keys 'uuid', 'version' and 'face' is returned.
+    Otherwise, an empty dictionary is returned.
+    """
+
+    obj = {}
+    urlinfo = urlsplit(picurl)
+    if urlinfo.netloc == 'c1.scryfall.com':
+        parts = urlinfo.path.split('/')
+        obj['version'] = parts[3]
+        obj['face'] = parts[4]
+        obj['uuid'] = parts[-1].rsplit('.')[0]
+    elif urlinfo.netloc == 'cards.scryfall.io':
+        parts = urlinfo.path.split('/')
+        obj['version'] = parts[1]
+        obj['face'] = parts[2]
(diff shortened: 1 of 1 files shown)


refresh.py
```diff
@@ -0,0 +1,180 @@
...
+            yield from list_obj['data']
+
+def parse_picurl(picurl):
+    """
+    Parse a Scryfall picURL into its components.
+
+    The Scryfall picURL must be in one of those forms:
+
+        - https://c1.scryfall.com/file/scryfall-cards/<version>/<face>/*/*/<uuid>.jpg
+        - https://cards.scryfall.io/<version>/<face>/*/*/<uuid>.jpg
```
tooomm, line 64: Do we need to account for any of the other variants?
https://scryfall.com/blog/upcoming-api-changes-to-scryfall-image-uris-and-download-uris-224

Elarnon: Not sure. It looks like only cards.scryfall.io is used for cards currently.

Maybe it would make sense for the script to check that it can re-parse the resolved URL and print out an error otherwise? This would indicate a change in scryfall's URLs that needs to be reflected in the script.

ZeldaZach:
I'd be okay with something like this, but it would need to be updated to address conflicts, and potentially apart of the CLI GHA

Elarnon:
Rebased the PR. I have also separate the commit introducing the script from the actual update to `tokens.xml` to make it slightly easier to work with.

> apart of the CLI GHA

Can you point me to the file where this should happen?

tooomm:
>Can you point me to the file where this should happen?

Not exactly sure, but I guess Zach is suggesting to have a GitHub action workflow that runs the script on a schedule to keep the links updated.
Such configuration files are placed in `.github/workflows`.
If you need help, let me know. I can also look into that.

<br>

>but it would need to be updated to address conflicts

Elarnon:
> > Can you point me to the file where this should happen?
> 
> Not exactly sure, but I guess Zach is suggesting to have a GitHub action workflow that runs the script on a schedule to keep the links updated. Such configuration files are placed in `.github/workflows`. If you need help, let me know. I can also look into that.
> 

Ah, I thought from the previous message that there was a specific workflow that was meant to be added to. I can set up a workflow to run the script on a schedule, but I would need to know when and how the script must run — I don't think that this is my call to make. I also think that having the script w/o the automation is a valuable first step and that automation could happen separately (when someone has the time to do it), but that is not my call to make either.

> > but it would need to be updated to address conflicts

I did this last time I commented, but then the `tokens.xml` file on `master` was changed :) I have just reran the script again, but it makes little sense for me to keep doing it if the PR is not going to be merged.

tooomm:
>I also think that having the script w/o the automation is a valuable first step and that automation could happen separately (when someone has the time to do it), but that is not my call to make either.

Sure! We can look into it in a second step.

tooomm:
>>but it would need to be updated to address conflicts
>
>I did this last time I commented, but then the tokens.xml file on master was changed :) I have just reran the script again, but it makes little sense for me to keep doing it if the PR is not going to be merged.

Oh sorry, I misunderstood the "conflicts" part I guess.

I was referring to https://github.com/Cockatrice/Magic-Token/pull/209#pullrequestreview-1336008804:
>Maybe it would make sense for the script to check that it can re-parse the resolved URL and print out an error otherwise? This would indicate a change in scryfall's URLs that needs to be reflected in the script.

Do you still want to look into that and think it's helpful?

<br>

Also, do you mind removing the changes of the scrip run from the PR to have it cleaner?
That keep this PR about the script and not the changes it'll do. That way, we also avoids conflicts in case the token file changes in between again, as you realized code reviews take a while and people only pop in once in a while...

Elarnon:
> I was referring to [#209 (review)](https://github.com/Cockatrice/Magic-Token/pull/209#pullrequestreview-1336008804):
> 
> > Maybe it would make sense for the script to check that it can re-parse the resolved URL and print out an error otherwise? This would indicate a change in scryfall's URLs that needs to be reflected in the script.
> 
> Do you still want to look into that and think it's helpful?

This is easy to do, I only did not do it because there was no acknowledgement of the suggestion. I have added it to the script (see [here](https://github.com/Cockatrice/Magic-Token/pull/209/commits/e9a7dacb4809e39f6d399d7b06ea1f564ad6bb0a#diff-639351c5127d0d05ff20715dd702399aa5d0124ac4a36a0980f8404f2c87025cR110-R115)).
 
> Also, do you mind removing the changes of the scrip run from the PR to have it cleaner? That keep this PR about the script and not the changes it'll do. That way, we also avoid conflicts in case the token file changes in between again - as you realized, code reviews take a while and people only pop in once in a while...

I have removed the changes to `tokens.xml` from the PR and opened #249 with only the `tokens.xml` changes (I will not keep updating that PR, but at least it can be merged separately without having to review the script if wanted; feel free to close it otherwise).

tooomm:
Thank you!
I'll leave this a few more days open for @ZeldaZach or @ebbit1q to check and comment.

>This is easy to do, I only did not do it because there was no acknowledgement of the suggestion. I have added it to the script (see [here](https://github.com/Cockatrice/Magic-Token/pull/209/commits/e9a7dacb4809e39f6d399d7b06ea1f564ad6bb0a#diff-639351c5127d0d05ff20715dd702399aa5d0124ac4a36a0980f8404f2c87025cR110-R115)).

It would be easier to spot and review this change/addition if you not force pushed and overwrote the commit history. ;)
We will squash all commits on merge anyway.

Elarnon:
> It would be easier to spot and review this change/addition if you not force pushed and overwrote the commit history. ;) We will squash all commits on merge anyway.

You can use the "Compare" button to see the changes:
image

Although you are right, it probably should have been a separate commit; I was already rebasing to remove the changes to the `tokens.xml` (got scared by #220) and didn't think too much about it, sorry!

tooomm:
I think we can move on here as there are no further comments or request.

Will open a PR for the CI automation on a reasonable schedule which takes care of creating PRs to keep the image links up to date.

PR #210, opened 2023-03-18
Remove '(Token)' from Some Token Entries

SlightlyCircuitous:
*Removes (Token) from 37 entries (mainly Embalm/Enternalize tokens) to bring them in line with Wizards/Scryfall token naming conventions 

*Tokens made by Squad are already named without (Token) so it also unifies the naming scheme in the file 

*Leaves a space after each name to differentiate it from its card counterpart

There is a short conversation on the Discord #dev channel from 9/27/22 related to this PR.

### DIFF
--- a/tokens.xml
+++ b/tokens.xml
@@ -31,7 +31,7 @@
         </card>
          -->
         <card>
-            <name>Adorned Pouncer (Token)</name>
+            <name>Adorned Pouncer </name>
             <text>Double strike</text>
             <prop>
                 <colors>B</colors>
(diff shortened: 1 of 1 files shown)


SlightlyCircuitous:
So, would you prefer just removing (Token) from the Embalm/Enternalize tokens and leaving the rest alone? I would at least like them to be consistent with Squad.

SlightlyCircuitous:
Happy to do that too; all I really want is a consistent naming convention so I know what to do when adding new tokens.

ebbit1q:
we could ask for some input from the community on this

ebbit1q:
I did an automated check of how many cards would be affected:
- there are 37 cards that currently have the suffix (Token) all of these have a card in the cards.xml that is the same
- there are 8 cards that currently just have a space to differentiate them from cards in the cards.xml, these are the cards with the squad ability but also Ornithopter and Storm Crow.

SlightlyCircuitous:
Seems like we should just change the 8 to go with the majority then, but we'll see what the community says

ebbit1q:
yes, it seems changing the squad cards to match has the preference.

SlightlyCircuitous:
Ok, I will close this in favor of doing that.

PR #212, opened 2023-03-20
Move CLB Angel Warrior art to Proper Entry

SlightlyCircuitous:
*Moves CLB Angel Warrior token (4/4 W Flying, Vigilance) to 4/4 W Flying, Vigilance entry instead of the 4/4 W Flying entry


tooomm (approved):
👍

Issue #214, opened 2023-03-31
Kibo's Bananas


Mortymacabre:
is there any way to add an art for banana tokens i understand  wizards never made one but people dont read often enough that its becoming an issue playing he deck 
![29125](https://user-images.githubusercontent.com/129471040/229002883-1bc48afd-71ec-4b0f-8ea6-cf72972c0c53.jpg)

ebbit1q:
the artist of this isn't properly credited, using google lens I found it is made by jackson moyer: https://www.artstation.com/artwork/NGaXEd

if you were to add their name to the bottom of your card we can include the url to it in the xml here: https://github.com/Cockatrice/Magic-Token/blob/6e1ce596bb498a7aecd1b49f544cf0045acb293f/tokens.xml#L507

SlightlyCircuitous:
Can we assume fair use with attribution here? I didn't see anything about license either way on that site.

SlightlyCircuitous:
I had a go at a token using mtgdesign to get a token frame. Please let me know if it's sufficient.

ebbit1q:
fair use with attribution is generally accepted yeah, if the author were to say we've misappropriated their art we'd remove it of course.

SlightlyCircuitous:
fair enough.

PR #215, opened 2023-03-31
Add image to banana token

SlightlyCircuitous:
Fixes https://github.com/Cockatrice/Magic-Token/issues/214

The picture used for the card is a public domain picture of a bunch of bananas with a GIMP olify filter applied. I have credited the person who uploaded the picture to Wikimedia Commons anyway.

### DIFF
--- a/tokens.xml
+++ b/tokens.xml
@@ -504,7 +504,7 @@ Creature tokens you control have flying and vigilance.</text>
                 <maintype>Artifact</maintype>
                 <cmc>0</cmc>
             </prop>
-            <set>J22</set>
+            <set picURL="https://github.com/SlightlyCircuitous/image-storage/blob/main/Banana.jpeg?raw=true">J22</set>
             <reverse-related>Kibo, Uktabi Prince</reverse-related>
             <token>1</token>
             <tablerow>1</tablerow>


Mortymacabre:
ive reran oracle and it still isnt popping up

SlightlyCircuitous:
well, this hasn't been merged into the client yet, so it won't show up until that happens. You should be able to see the link under 'Files Changed' and look at the image that way.

Mortymacabre:
Just curious, when it will be merged with the client? I don't know how long these kinda things take. I really appreciate you guys getting this done

SlightlyCircuitous:
One of the maintainers has to approve it first. Probably will happen in the next few days unless something is seriously wrong with it.

ebbit1q (approved):
works fine, don't really like the way it splits the line right before {G} though.

SlightlyCircuitous:
I could probably fix that by forcing a line break somewhere else. Is there a place for the line break you would prefer?

ebbit1q:
decrease the font size by a small amount and it'll have this effect.

SlightlyCircuitous:
ah, yeah, good point.

SlightlyCircuitous:
I fixed the image and re-uploaded it to my image-storage repo. The link is the same so it should just work in Cockatrice (it does for me, anyway).

ebbit1q:
it does after you remove it from the image cache yeah

PR #216, opened 2023-03-31
Add Otharri, Suns' Glory to Experience Counter

SlightlyCircuitous:


tooomm (approved):
Looks about right, thanks SlightlyCircuitous!

tooomm:
See #218, had to recreate the PR as I could not fix it when adding the date and other merges happened in parallel. There were also commits with old changes on this PR branch.

Recreating this simple change again in a clean way was the easiest.

ebbit1q:
I don't understand, why not rebase it?

PR #217, opened 2023-04-01
Add image to J22 Pirate token

SlightlyCircuitous:
Art is from _Pirates_, an early 1800's watercolor by John Masey Wright. More progress towards issue https://github.com/Cockatrice/Magic-Token/issues/85.


ebbit1q (approved):
thanks

PR #218, opened 2023-04-01
Add Otharri, Suns' Glory to Experience Counter

tooomm:
This re-adds [`4e5bc82` (#216)](https://github.com/Cockatrice/Magic-Token/pull/216/commits/4e5bc82dc91c44b79fad6ea13552f7b9acdd9617) by @SlightlyCircuitous


ebbit1q:
you could have used the workflow to update the version here, the workflow goes from 20230401 to 20230401b btw, we never make an a version

PR #219, opened 2023-04-01
Add (Token) to Certain Token Names

SlightlyCircuitous:
Based on the discussion in https://github.com/Cockatrice/Magic-Token/pull/210 and a beta-tester straw poll, this PR adds (Token) to the names of tokens created by Squad as well as the Ornithopter Token and the Storm Crow token to bring them in line with how the other named tokens are differentiated from their same-name card counterparts.


ebbit1q:
you forgot to update the version @ZeldaZach

PR #222, opened 2023-04-10
Add MOM Tokens

SlightlyCircuitous:
*Added entries for 13 new tokens
*Added set lines for 8 reprinted tokens
*Reverse-related an existing Knight token to Elspeth's Talent

I elected not to add the art link for the Dinosaur token as the actual token card does not have Trample printed on it and Ghalta and Mavren make an X/X Dinosaur with Trample. I am happy to include it if we feel like its better to have a close-enough token than nothing (or if WotC is now treating Trample like Haste for some reason). Just let me know.

New cards Cutthroat Negotiator and The Western Cloud make Treasure tokens that come into play tapped; is there any way to enforce this in the XML or is that only possible for cards?

### DIFF
--- a/tokens.xml
+++ b/tokens.xml
@@ -2186,6 +2186,21 @@ When this creature dies, create a colorless Treasure artifact token.</text>
             <tablerow>2</tablerow>
         </card>
         <card>
+            <name>Dinosaur Token  </name>
+            <text>Trample</text>
+            <prop>
+                <colors>G</colors>
+                <type>Token Creature — Dinosaur</type>
+                <maintype>Creature</maintype>
+                <cmc>0</cmc>
+                <pt>*/*</pt>
+            </prop>
+            <set>MOM</set>
+            <reverse-related>Ghalta and Mavren</reverse-related>
+            <token>1</token>
+            <tablerow>2</tablerow>
+        </card>
+        <card>
             <name>Djinn Monk Token</name>
             <text>Flying</text>
             <prop>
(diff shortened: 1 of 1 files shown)


ebbit1q:
I'm not sure what you mean with "only possible for cards", such a distinction does not exist in the xml. Tokens are cards, cards can have the cipt property, relations can't tell the card to come into play tapped. You should only use the cipt property if the card always comes into play tapped, ie it has this comes into play tapped printed on it. This is also what oracle does when it creates the cards.xml.

ebbit1q:
one thing you might not be aware of about blightsower thallid: normally when a card transforms you'll want to mark its tokens as "exclude" in order to not have it accidentally make a token when you transform it, in this case not marking it exclude is a good thing to do intentionally as it explicitly only makes its token when it transforms.

ebbit1q (approved):

SlightlyCircuitous:
OK, good to know. I will have the MOC tokens up shortly and I will go back and make sure I do that for them.

ebbit1q:
Yeah, I just did a check and noticed we're not very consistent with this.

ebbit1q:
Ok, apparently it works differently than expected, tokens will always take precedence over any transform ability, the default token action will only transform the card when it has no token relations even when they are marked as exclude. This is not ideal but not unreasonable.

We should not mark all tokens as exclude on transforming cards, it just doesn't have an effect.

PR #223, opened 2023-04-11
Add MOC Tokens

SlightlyCircuitous:
*Added entries for 3 new tokens
*Added set lines for 40 reprinted tokens
(I replaced the butterfly token art instead of merely adding the new art as the existing token art does not use a format consistent with any printing of tokens, but I can put it back if we want to keep it)
~~*Corrected several instances of "exlude"~~ These were removed in a subsequent commit
~~*Excluded tokens from The Argent Etchings to prevent tokens being made on transform~~ This doesn't do anything
*Removed instances of exclude on tokens made by transforming MOM cards as they don't do anything (See https://github.com/Cockatrice/Magic-Token/pull/222#issuecomment-1503739091)


ebbit1q:
`Goro-Goro and Satoru` and `Slimefoot and Squee` are missing their token entries.

ebbit1q (approved):
awesome, thanks

PR #224, opened 2023-04-19
Fix Transforming Tokens

SlightlyCircuitous:
*Adds attach="transform" to Incubator Token, Day, and Night to allow tokens to transform using the context menu


ebbit1q (approved):
thanks

Issue #190, opened 2022-12-02
Duplication of challenge set cards between cards.xml and challenge_tokens.xml


SlightlyCircuitous:
All of the non-creature spells in the challenge sets (and the Garruk the Slayer promo) are present in both the cards.xml file and the challenge_tokens.xml file. In the cards.xml file, they are filed under different set names (TFTH, TBTH, TDAG, THP1, THP2, THP3, and PPC1). This means that each comes up as two different cards if you use the challenge tokens file alongside the cards.xml file.

Would it make sense to add the challenge set creature tokens to tokens.xml using the same sets for them as the cards.xml file does and then get rid of the challenge_tokens.xml file entirely? Or possibly just remove the non-creature spells from challenge_tokens.xml and rename the sets there to match cards.xml? I'm happy to do either if anyone feels it's warranted.

PR #225, opened 2023-05-04
Merge Theros Challenge Deck creatures into token.xml

SlightlyCircuitous:
Fixes https://github.com/Cockatrice/Magic-Token/issues/190.

Every non-creature spell in the Theros Challenge Decks (as well as Garruk the Slayer) is currently listed in cards.xml and displays in Cockatrice by default. This means that if you use challenge_tokens.xml to have access to the creature spells in a given challenge deck, you will have duplicated entries for most of the cards in those decks. Further, all the spells are listed in different sets in cards.xml and challenge_tokens.xml, further increasing confusion. One option is merely to remove the non-creature spells from challenge_tokens.xml, but it doesn't really make sense to maintain a separate xml file for just the 17 creature spells.

Instead, this PR would merge the creature spells into tokens.xml, similar to how the non-creature spells are in cards.xml, making the full challenge decks available by default. It would also bring the set of each creature spell in line with its set in cards.xml.


README.md
```diff
@@ -6,6 +6,4 @@ Magic-Token [![](https://img.shields.io/badge/dynamic/xml.svg?label=version&url=
 This repo contains token information in [Cockatrice](https://github.com/cockatrice/cockatrice)'s [XML card database format](https://github.com/Cockatrice/Cockatrice/wiki/Custom-Cards-&-Sets#to-add-your-own-custom-cards-follow-these-steps) for Magic: The Gathering.<br />
 It describes [![](https://img.shields.io/badge/dynamic/xml.svg?label=&colorB=white&url=https%3A%2F%2Fraw.githubusercontent.com%2FCockatrice%2FMagic-Token%2Fmaster%2Ftokens.xml&query=count(%2F%2Fcard))](https://github.com/Cockatrice/Magic-Token/blob/master/tokens.xml) individual tokens with linked art and the relation to their creators.
 
-For normal tokens, save the file found here: [**tokens.xml**](https://raw.githubusercontent.com/Cockatrice/Magic-Token/master/tokens.xml)
-
-For tokens from the Theros block challege decks, save the file found here: [**challenge_tokens.xml**](https://raw.githubusercontent.com/Cockatrice/Magic-Token/master/challenge_tokens.xml)
+A copy of the tokens file included with Cockatrice can be found here: [**tokens.xml**](https://raw.githubusercontent.com/Cockatrice/Magic-Token/master/tokens.xml)
```
tooomm, line 9:
```suggestion
Cockatrice pulls its copy of the token file from here: [**tokens.xml**](https://raw.githubusercontent.com/Cockatrice/Magic-Token/master/tokens.xml)
```

A bit shorter version to make it fit a single line.

tooomm (commented):
I did fix the CI run and included a wording suggestion.

Had no time to look at and compare the actual file merge.

But for a better understanding:
You only moved the 17 creature spells to the tokens.xml file. All non-creature spells are not moved and are no longer part of this repo. We rely on the cards.xml content to make those available via special sets (e.g. TFTH, TBTH, TDAG, THP1...) which include them - but not the creature spells which you moved and preserved in the tokens.xml file.

README.md
SlightlyCircuitous: Thank you for updating the CI check; I was a little worried that that failing check would scare reviewers off, but I wasn't sure how to (or if I even could) change it myself. 

And, yes, I only moved the creature spells as the rest are in cards.xml already. It seems odd to have them in two places (either in both cards.xml and tokens.xml, or in both cards.xml and challenge_tokens.xml) since that creates redundant entries in the client. However, I could understand not wanting to give up control of the non-creature spells and leave them in the hands of whatever processes create cards.xml. It would probably also be fine to put all challenge deck creature and non-creature spells in tokens.xml, but then the non-creature spells would have to be removed from cards.xml somehow, which I don't know how to do.

SlightlyCircuitous: Thank you for updating the CI check; I was a little worried that that failing check would scare reviewers off, but I wasn't sure how to (or if I even could) change it myself. 

And, yes, I only moved the creature spells as the rest are in cards.xml already. It seems odd to have them in two places (either in both cards.xml and tokens.xml, or in both cards.xml and challenge_tokens.xml) since that creates redundant entries in the client. However, I could understand not wanting to give up control of the non-creature spells and leave them in the hands of whatever processes create cards.xml. It would probably also be fine to put all challenge deck creature and non-creature spells in tokens.xml, but then the non-creature spells would have to be removed from cards.xml somehow, which I don't know how to do.

tooomm (approved):
I resolved conflicts and added a note.

Think this is good to go and a reasonable change. 👍

<br>

Changing the content of cards.xml is something that would need to be done in the Oracle database builder and our MTGJSON parser at Cockatrice.
Plan is still to get rid of Oracle at one point and provide the parsed files ready to use similar to Spoiler and Token.

SlightlyCircuitous:
Lovely, thank you. I think that note definitely helps since these really aren't tokens persay.

Issue #226, opened 2023-06-14
Lord of the Rings: Tales of Middle Earth Tokens


Keither85247:
Hey,

I am dying to play this set with my friends. Will anyone take the time to upload the tokens and relate them to the cards? Thank you very much.

ebbit1q:
We rely on volunteers to contribute them, you too can be a contributor. You can edit the tokens.xml in a text editor and upload them in a pull request, you can take a look at other prs as an example. eg https://github.com/Cockatrice/Magic-Token/pull/223/files

ebbit1q:
We have a complete reference for the properties of the xml here https://github.com/Cockatrice/Cockatrice/wiki/Custom-Cards-&-Sets#to-add-your-own-custom-cards-follow-these-steps

We don't have like a guide on how to specifically contribute to the tokens.xml, it comes down to adding `<reverse-related>` tags to tokens that are already in the list and adding new `<card>` entries with the right properties when a token is unique. Also you can add art (scryfall scans are preferred) with the `<set>` tags.

After editing you can use an xml validator to check if your file is correctly formatted (no missing closing tags etc). Then you can create a pr on github: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/creating-a-pull-request

If you'd like to ask some questions directly you can visit the #dev channel on our discord: https://cockatrice.us/discord

SlightlyCircuitous:
I could write up the process I follow/notes I have into a guide if there is interest in that sort of thing.

Keither85247:
That would be great. I would definitely find that beneficial. Thank you!

From: SlightlyCircuitous ***@***.***>
Date: Thursday, June 15, 2023 at 6:29 PM
To: Cockatrice/Magic-Token ***@***.***>
Cc: Keither85247 ***@***.***>, Author ***@***.***>
Subject: Re: [Cockatrice/Magic-Token] Lord of the Rings: Tales of Middle Earth Tokens (Issue #226)

I could write up the process I follow/notes I have into a guide if there is interest in that sort of thing.

SlightlyCircuitous:
[...]

PR #227, opened 2023-06-15
Add LTR and LTC Tokens

SlightlyCircuitous:
*Fixes https://github.com/Cockatrice/Magic-Token/issues/226
*Adds entries for 5 new tokens and set lines for 5 reprints from LTR 
*Adds entries for 5 new tokens and set lines for 10 reprints from LTC

I currently have "The Ring" listed with type Emblem and have it reverse related to any card that says "The ring tempts you." as its referred to as an emblem on the "The Ring Tempts You" reminder card. I have listed said reminder card with type Token and related it to "The Ring" so that "The Ring" can make the reminder card if needed. Please feel free to change the types around if you think something fits better.


ebbit1q:
has the entire set been spoiled yet?

SlightlyCircuitous:
As far as I can tell, yes. All the tokens that we expect to have are up, the card counts line up with the expected number of cards on the wiki, and Scryfall has both LTR and LTC listed on the front page as new additions (granted, they don't say 'full preview' though).

SlightlyCircuitous:
Also, something I just noticed while looking at the wiki is that the Halfling token has printed type 'Tolkien Creature' which is how it is listed on Scryfall (doesn't show up in a search for t:token). My script picked that up when making the entry, but I assume we would want the regular "Token Creature" in the xml.

tokens.xml
```diff
@@ -10024,11 +10243,13 @@ This creature can't be enchanted.</text>
             <reverse-related count="x">Marut</reverse-related>
             <reverse-related count="x">Master of Ceremonies</reverse-related>
             <reverse-related>Meet and Greet "Sisay"</reverse-related>
+            <reverse-related count="2">Mines of Moria </reverse-related>
```
ebbit1q, line 10246: why is there a space at the end?

ebbit1q (commented):
Thank you for adding these!

PR #229, opened 2023-06-22
Add ARC Scheme Tokens

SlightlyCircuitous:
*Progress towards https://github.com/Cockatrice/Magic-Token/issues/85
*Adds a 4/6 golem token and a 3/3 elemental token for The Iron Guardian Stirs and Evil Comes to Fruition, respectively

I looked high and low for some good public domain art to make these tokens with, but I could not find anything suitable, so both these tokens merely use art from their respective cards. I don't know if this is actually a problem for anyone but me, so I figured I would just make this PR and see if anyone raises any objections.

### DIFF
--- a/tokens.xml
+++ b/tokens.xml
@@ -3219,7 +3219,7 @@ When Earthshaker Khenra enters the battlefield, target creature with power less 
                 <cmc>0</cmc>
                 <pt>3/3</pt>
             </prop>
-            <set>ARC</set>
+            <set picURL="https://github.com/SlightlyCircuitous/image-storage/blob/main/Elemental.jpeg?raw=true">ARC</set>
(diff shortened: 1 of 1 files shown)


ebbit1q:
I doubt many people are playing archenemy on cockatrice but these look neat. Considering the art is already on the cards themselves and is attributed I see no problem. Worst case your repo gets asked to remove the picture, linking the picture in this repo would always be fine.

SlightlyCircuitous:
Yeah, I assume Archenemy is not a great format on Cockatrice since you'd need to work around the extra deck, but I love to check boxes and issue 85 is always calling to me.

SlightlyCircuitous:
Is this OK to merge?

PR #231, opened 2023-07-01
Fix treasure token name

### LINKED ISSUE #230: The Undercity should be able to create a Treasure Token
As the title suggests the Undercity can create the 4/1 Skeleton Token from "Catacombs" but can't create the Treasure Token from "Stash"

image

tooomm:
Fix #230


ebbit1q (approved):
looks good

PR #233, opened 2023-07-26
Add CMM Tokens

SlightlyCircuitous:
*Adds entries for 8 new tokens
*Adds set lines for 69 reprints
*Splits up entries for 3/1 R Elemental and 3/1 R Elemental with Haste and entries for 1/1 R Elemental and 1/1 R Elemental with Haste as WotC has decided to print those tokens with Haste again and not all the reverse-related cards make tokens with Haste 
*Changes text for Chandra, Torch of Defiance to reflect new oracle text 
*Changes reverse-related for Lightning Coils to count="x=5" since it should always make at least 5 tokens when it goes off

Please let me know if any of these changes are too far and I will revert them.


ebbit1q:
I would prefer it if in the future you split up your prs, one for the new set and one for the changes, this makes reviewing easier.

ebbit1q:
thanks again for helping out, if you notice I've not merged your pr for a week again try pinging me on discord.

ebbit1q:
use your own judgement in that case, if the two prs interfere with each other it's possible to have them be dependent on each other, alternatively simply split up the changes into multiple commits on the same pr for reviewing, if you think it's not worth the effort then simply don't worry about it.

PR #242, opened 2023-11-04
Add LCI and LCC Tokens

SlightlyCircuitous:
*Adds 10 new tokens
*Adds 28 set lines for reprinted tokens
*Relates all new and reprinted tokens to new cards in LCI and LCC 
*Adjusts Oracle text for tokens with new reminder text


ebbit1q:
I've corrected some things, please check if I didn't miss anything and I will merge this

SlightlyCircuitous:
Thanks. I just did not get alphabetization right this time, weird.

Looking over the changes, I thought that transform cards didn't have to have their tokens excluded since transform took precedence as the default token action. Or were they excluded for a different reason?

ebbit1q (approved):
many thanks

ebbit1q:
I think it just makes sense to exclude the transform cards

PR #244, opened 2023-11-12
Clean up elements

SlightlyCircuitous:
Based on the conversation in https://github.com/Cockatrice/Cockatrice/pull/4933, I have removed the second <cards> element and added some comments denoting different subgroups of reminder tokens, as suggested in https://github.com/Cockatrice/Cockatrice/pull/4933#issuecomment-1793816240.


ebbit1q (approved):
thanks, also great to have the added comments to the other sections

tooomm:
Quite some sub-groups there. Thanks for highlighting and explaining them with the new comments! :)

How do you actually validate the file token file?
Which tool are you using for the job?

SlightlyCircuitous:
I use `xmllint` in the command line with cardsv4.xsd as the schema

PR #249, opened 2023-12-17
Migrate c1.scryfall.com links to cards.scryfall.io

Elarnon:
This is the result of running the script from #209, without the script itself.


tooomm:
Superseded by #261

PR #255, opened 2024-01-30
Relate new CLU cards to exisiting tokens

SlightlyCircuitous:
*Adds reverse-relations to ~~9~~ 22 new cards released in CLU that make existing tokens.

There were no tokens printed in CLU so no new art additions are necessary. New tokens from MKM/MKC to follow in a separate PR.


SlightlyCircuitous:
Realized that I missed several cards in my scryfall search. Converting to a draft while I add those in.

SlightlyCircuitous:
Should have them all now. Those keywords always trip me up.

tokens.xml
```diff
@@ -1934,16 +1936,21 @@ Cloud Sprite can block only creatures with flying.</text>
             <reverse-related>Angelic Sleuth</reverse-related>
             <reverse-related count="x" exclude="exclude">Angelic Sleuth</reverse-related>
             <reverse-related>Antarctic Research Base</reverse-related>
+            <reverse-related>Ballroom<reverse-related>
```
ebbit1q, line 1939: typo
```suggestion
            <reverse-related>Ballroom</reverse-related>
```

tokens.xml
```diff
@@ -10956,6 +10979,8 @@ This creature can't be enchanted.</text>
             <reverse-related>Contested Game Ball</reverse-related>
             <reverse-related count="2">Contract Killing</reverse-related>
             <reverse-related>Corsair Captain</reverse-related>
+            <reverse-related>Covetous Elegy</reverse-related>
```
ebbit1q, line 10982: I think this should always be x

SlightlyCircuitous: My thought process was that in a two-player game, you are making 1 or 2 tokens since your opponent saved 1 or 2 creatures. In that instance, you might just click the make token option 1 or 2 times, possibly making the x option more trouble that it's worth. But maybe I am overthinking it and just having an x option for making an always unknown amount of tokens is the better idea. I am fine with that change.

PR #258, opened 2024-02-03
Add picture to UNF Sketch token

SlightlyCircuitous:
*Adds a "blank canvas" token frame to reflect the drawn nature of the token

*Adjusts the reminder text to focus on where the art should come from rather than replicating the exact effect wording from D00-DL, Caricaturist (also the longer version requires much smaller text to fit in the text box, which looked silly)

Progress on https://github.com/Cockatrice/Magic-Token/issues/85, though Sketch is not mentioned by name in that issue.

I am open to suggestions for any changes to the art or reminder text that might better reflect the flavor of the token or enhance understanding.


SlightlyCircuitous:
Yeah, fair. It just feels better than a blank space to me, I guess.

Issue #262, opened 2024-03-02
CI: Update version workflow contains deprecated parts


tooomm:
See e.g. here:
https://github.com/Cockatrice/Magic-Token/actions/runs/8122265823

>[update_version](https://github.com/Cockatrice/Magic-Token/actions/runs/8122265823/job/22201519206#step:6:28)
The `set-output` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

PR #264, opened 2024-03-10
Replace set-output in updateversion.yml

SlightlyCircuitous:
- Fixes https://github.com/Cockatrice/Magic-Token/issues/262
- Updated the lines that use `set-output` in the beginning and in the Bash shell based on [Github's instructions ](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/)
- Updated the line that uses `set-output` in the Python shell based on [this discussion](https://github.com/orgs/community/discussions/28146)

I'm not really sure how to test changes to Github Actions, so please let me know what I need to do to make sure the new code works.

### DIFF
--- a/.github/workflows/updateversion.yml
+++ b/.github/workflows/updateversion.yml
@@ -101,6 +101,7 @@ jobs:
         id: version
         shell: python
         run: |
+          import os
           import re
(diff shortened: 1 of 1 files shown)


tooomm (approved):
Test

PR #265, opened 2024-04-06
Add OTC, OTJ, BIG Tokens

SlightlyCircuitous:
Changes:

* Adds 26 new token entries across three sets.
* Adds set lines and reverse relations for 42 reprinted tokens

Thoughts:

BIG is included here as it releases on the same day as OTJ and OTC and was originally meant to be an Aftermath set (like March of the Machine: The Aftermath) so it's very related.

For the Bounties, I gave them each `Bounty` for type and maintype, since that's how I've seen them referred to in WotC articles, and I related each of them to a single Wanted! entry so they can each make a copy of their rules reminder back face (it's the same for each Bounty). I wasn't sure if it was necessary to relate Wanted! to all the Bounties to let players create each Bounty from one copy of the reminder text card, but that's easily done if anyone thinks its useful. 

I modeled the reverse-relations for Plot after Foretell, since it's basically Foretell. However, I noticed that On An Adventure (which is also similar to both) uses `attach="attach"` whereas Foretell doesn't. Is this an inconsistency or does the special nature of the split Adventure cards necessitate `attach` in a way that Foretell (and Plot) doesn't?

### DIFF
--- a/tokens.xml
+++ b/tokens.xml
@@ -255,6 +255,7 @@ When Angel of Sanctions enters the battlefield, you may exile target nonland per
                 <cmc>0</cmc>
                 <pt>4/4</pt>
             </prop>
+            <set picURL="https://cards.scryfall.io/large/front/b/1/b1b0c498-b457-4865-96da-a69647179456.jpg?1712319499">OTC</set>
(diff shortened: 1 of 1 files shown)


ZeldaZach:
Probably an inconsistency, we can fix it at some point

PR #269, opened 2024-06-14
Add MH3 and M3C Tokens

SlightlyCircuitous:
- Add Modern Horizons and Modern Horizons Commander tokens 
- Update Shapeshifter reminder text to match new printing

Given that the new Spellgorger Weird token has a mana cost of 2R and is seemingly a token copy of an existing card of the same name, should I adjust the token's cmc to 3 and put "(Token)" after the token name like with Eternalize/Embalm cards?


ebbit1q:
yes, we'd like all tokens that have a similar name to a real card to have (Token) after it

tokens.xml
```diff
@@ -1629,6 +1654,22 @@ Whenever this creature attacks, target attacking creature gains flying until end
             <token>1</token>
             <tablerow>2</tablerow>
         </card>
+        <card>
+            <name>Cat Warrior Token</name>
```
ebbit1q, line 1658: this is a duplicate of an existing token, it is sorted alphabetically, ie W after T

SlightlyCircuitous:
Thanks for the fixes!

ebbit1q:
you're welcome

PR #272, opened 2024-07-04
Add ACR tokens

SlightlyCircuitous:
Changes:
* Adds token entries for 4 new tokens
* Adds set lines and reverse-relations for 4 existing tokens

~~_A Mysterious Creature_ does not seem to have been reprinted (or, at least, it hasn't made it to Scryfall), but 4 tokens in this set have Disguise or Cloak so those relations have been included in this PR anyway.~~ _A Mysterious Creature_ has since been added to Scryfall.

I have tried to be as inclusive as possible for the multi-part effect of _Origin of the Hidden Ones_, but I am happy to pare down those options if we think it's too much.


ebbit1q (approved):
thank you!

PR #276, opened 2024-07-24
Add BLB and BLC Tokens

SlightlyCircuitous:
* Adds new token entries and reprinted token art from BLB and BLC sets
* Adds reminder text to the Squid token to be in line with current oracle text


thopp-tudelft (approved):
excellent, thanks

ebbit1q:
yay

PR #277, opened 2024-07-27
Remove duplicate elemental token entry

SlightlyCircuitous:
There were two entries for the 4/4 W elemental token created by _Hoofprints of the Stag_, one containing the BLC, C16, and LRW art and one containing just the C20 art. I have combined them and adjusted the spaces on the entries following it as needed.


ebbit1q:
please don't adjust the following entries, people with mismatching database xmls will run into issues, we should allow people to update at their own pace

SlightlyCircuitous:
Good point about mismatched entries, I hadn't considered that. I have un-adjusted the entries following the duplicated entry.

ebbit1q (approved):
thanks for the fix

PR #278, opened 2024-07-29
Add cmc and manacost to 40K Squad tokens

SlightlyCircuitous:
Since the `Squad` keyword makes an exact copy of a card, any token made with `Squad` should have a cmc and mana cost equal to its parent card. While there are cards with `Squad` in a few sets (40K, LTC, and PIP), only the 40K cards come with printed tokens that are in `tokens.xml`. I have updated those tokens with appropriate cmc and mana costs.

Both `Embalm` and `Eternalize` also make copies of the parent card and have printed tokens in `tokens.xml`, but both keywords specify that the resulting token has a mana cost of 0 so no change is necessary there.


ebbit1q (commented):
nice catch, we probably didn't notice the small difference at the time

PR #280, opened 2024-08-29
Add missing relation for Ral, Crackling Wit

tooomm:


ebbit1q:
instead of merging immediately we have the workflow that updates the version on the pr with an approval

tooomm:
@ZeldaZach FYI
You just need to leave an approving review and wait a few seconds to get the version/dates updated automatically.

PR #283, opened 2024-09-24
DSK/DSC Tokens

celo0:


tokens.xml
```diff
@@ -1,10 +1,10 @@
-<?xml version="1.0" encoding="UTF-8"?>
+<?xml version='1.0' encoding='UTF-8'?>
```
ebbit1q, line 1: what's this change for?

tokens.xml
```diff
@@ -4339,6 +4394,19 @@ This creature's power and toughness are each equal to the number of instant and
             <token>1</token>
             <tablerow>1</tablerow>
         </card>
+        <card>
+            <name>Everywhere</name>
+            <text>This land is a Plains, Island, Swamp, Mountain, and Forest. ({T}: Add {W}, {U}, {B}, {R}, or {G}.)</text>
+            <prop>
+                <type>Token Land</type>
+                <maintype>Please edit manually</maintype>
```
ebbit1q, line 4402:
```suggestion
                <maintype>Land</maintype>
```

tokens.xml
celo0: I think it was my lxml lib version when i runned the script to grab the new tokens

ebbit1q (approved):
thank you for your help!

tokens.xml
SlightlyCircuitous: For whatever reason, lxml defaults to single quotes in the header (probably because Python makes no distinction between single and double quotes) and I can't figure out a way to change it. It messes around the order of attributes in <cockatrice_carddatabase> too.

celo0: Next time I will revert it back to double quotes and the correct order of the attributes as well

PR #284, opened 2024-11-10
Add FDN Tokens

SlightlyCircuitous:
Changes:

- Adds 3 new tokens and 1 new emblem from Foundations and their reverse relations
- Adds art for 29 reprints from Foundations and their reverse relations
- Updates the `text` field of the 1/1 white Cat token with lifelink to include new reminder text
- Adds `count="4"` to *Release the Dogs* to better reflect the card effect


tokens.xml
```diff
@@ -8364,6 +8417,21 @@ A card with morph can be turned face up any time for its morph cost.)</text>
...
+            <prop>
+                <colors>U</colors>
+                <type>Token Creature — Ninja</type>
+                <maintype>Creature</maintype>
+                <cmc>0</cmc>
+                <pt>2/1</pt>
+            </prop>
+            <set picURL="https://cards.scryfall.io/large/front/a/e/aeec04b1-475c-4e55-b72f-327ea5258146.jpg?1730485615">FDN</set>
+            <reverse-related>Kaito, Cunning Infiltrator</reverse-related>
+            <reverse-related>Kaito, Cunning Infiltrator Emblem</reverse-related>
```
ebbit1q, line 8431: we prefer related tags on the emblem for references in the same file

SlightlyCircuitous: gotcha. changed.

ebbit1q (approved):
thank you very much!

tokens.xml
```diff
@@ -1819,14 +1823,15 @@ Whenever this creature attacks, target attacking creature gains flying until end
         </card>
         <card>
             <name>Cat Token   </name>
-            <text>Lifelink</text>
+            <text>Lifelink (Damage dealt by this creature also causes you to gain that much life.)</text>
```
tooomm, line 1826: Hmm, I'm not so sure I like this addition here.

The M19 token had already a helper text printed on the token, did we had it in the file and removed it then when IKO came out and the printed token removed it as well? Just to stay in line? That sounds like a crazy maintenance work for no real benefit.
With the next reprint and no helper text, I doubt we remove it again to have the file in sync with the printed token.

I think the better approach is to uniformly not deal with any helper text, no matter what is printed on the token in a given or the newest edition, and stay simple and more edition agnostic. At least for basic keywords.

I'm aware that there is no strict guidance we follow right now, but seeing this made me think as I'm pretty sure we do not match the text on all newest prints probably (if that ever was the goal).

ebbit1q: it doesn't hurt to have the reminder text imo

it's not like there is too much text here, or that it has a meaningful impact on the file size

the reason why lifelink (or other evergreen keywords) only sometimes have reminder text is wizard's own policy, in order to make the game accessible to newcomers they make "core sets", core sets will have only existing keywords and get reminder text for evergreen keywords that in other sets is not needed, normal expansion sets are more oriented to more experienced players and will add new mechanics and keywords.

SlightlyCircuitous: The script I use to pull down tokens from Scryfall uses the oracle text (among other things) to differentiate between tokens that need novel entries and tokens that are reprints and just need a new set line.  This means that reminder text changes will be flagged as new tokens to the script. I updated the text here a little selfishly to keep up with the oracle text changes so my script can be more accurate. 

This, of course, relies on the assumption that once WoTC adds reminder text to the oracle text of a token, they won't take it away. If that's not true, then I guess the change is not useful and it doesn't matter.

ebbit1q: the reminder text will be there only on "core" sets: https://scryfall.com/search?as=grid&order=released&q=oracleid%3A65742c00-2809-4085-a950-8f93845c03d2+include%3Aextras&unique=prints

PR #287, opened 2024-11-12
Add J25 Token Relations

SlightlyCircuitous:
Changes:

- Adds 10 reverse-relations for first printings in Foundations Jumpstart that make existing tokens
- Adds 3 new token entries for first printings in Foundations Jumpstart that make novel tokens (no physical tokens have been printed in this set so no artwork is available currently)


ebbit1q (approved):
thanks!

PR #291, opened 2025-02-04
Add DFT and DRC Tokens

SlightlyCircuitous:
- Adds new token entries, reprinted art, and new reverse relations from Aetherdrift and Aetherdrift Commander sets
- Updates oracle text on Champion of Wits, Treasure, and Zombie tokens


ebbit1q (approved):
sorry about being so late reviewing this, please ping me on discord if I miss it again!

note that sld cards use universes within names (even though the text in oracle seems wrong)

PR #294, opened 2025-02-17
Separate DFT pilot token

SlightlyCircuitous:
The NEO pilot token and the DFT pilot token are not the same token, as I had previously assumed in https://github.com/Cockatrice/Magic-Token/pull/291. I can't find any evidence that the NEO pilot token has received errata to have the new _saddle_ effect that the DFT pilot token has, so they need to be two separate entries.


ebbit1q:
I think you're correct, these two tokens are separate

PR #298, opened 2025-04-01
Add TDM and TDC tokens

SlightlyCircuitous:
* Adds 9 new token entries and art for 41 reprinted tokens, as well as their reverse-relations
* Adds reminder text to the Experience counter token
* Updates the oracle text of Gold Token to be in line with new conventions for tokens referring to themselves


ebbit1q (approved):
thank you very much!

PR #299, opened 2025-04-04
Add missing Energy Reserve reverse-relations

SlightlyCircuitous:
Adds Energy Reserve reverse-relations to four cards for PIP and one from MH3 that generate energy counters

First reported on Discord


ebbit1q (approved):

ebbit1q:
oh wow, I fucked that one up

Issue #301, opened 2025-05-09
CRANK! Counter


MarsIsCoding:
![Image](https://github.com/user-attachments/assets/6aea9af9-eed6-474d-b4b8-d705af872783)

CRANK! Counter for Contraptions from the UST set. Not present in the app, essential to use those cards. Thanks for the amazing application!

PR #302, opened 2025-05-09
Add CRANK! state token

SlightlyCircuitous:
- Fixes https://github.com/Cockatrice/Magic-Token/issues/301
- Adds CRANK! state token for use with Contraptions
- Reverse-relates all Contraptions to CRANK! token

The text is sideways in the card image, which is a little disorienting, though I don't know if there is much to be done about that. Is it possible to display the token rotated 90° like the Siege cards or is that only supported for entries in the cards file?


SlightlyCircuitous:
As far as I can tell, image health check is failing here because Scryfall pulled the image/entry of a Spirit token in WOC for some reason. Link gives a 404 and I can't find it on their site. I can open another PR to remove it if that makes sense.

ebbit1q (approved):
thank you

ebbit1q:
I've added the landscapeorientation tags to it, like is on the battles
note that this only applies to new clients and old clients will display a warning when it finds thistag

SlightlyCircuitous:
Ah, amazing, thank you. I had meant to ask how to do that.

PR #303, opened 2025-05-18
Remove dead Scryfall links for unprinted WOC tokens

SlightlyCircuitous:
Four tokens from the Wilds of Eldraine commander set (WOC) are failing the picture health check. The URLs turn up 404 errors when navigated to in the browser and these tokens no longer appear on Scryfall's website when [searching for WOC tokens](https://scryfall.com/sets/twoc?as=grid&order=set). This PR removes the set lines/picURLs these tokens as they seem to not exist anymore.

I believe this has to do with how Scryfall handled the [physical misprint of the *Virture and Valor* deck](https://mtg.fandom.com/wiki/Wilds_of_Eldraine/Commander_decks#Tokens); they expected these tokens to exist based on previews, but the tokens were actually never printed and so were removed from Scryfall.

These tokens are: 
2/2 white Cat
3/3 green Elephant
1/1 BW Spirit with flying
1/1 green Human Monk with "{T}: Add {G}."


tooomm:
Thanks for the fix!

tooomm:
Looks like the currently still failing link is also related to that and got missed here?
Can you check & confirm, please @SlightlyCircuitous?

https://cards.scryfall.io/large/front/b/9/b93ed798-245a-46d6-8964-1aebc290008d.jpg?1692830136

PR #305, opened 2025-05-30
Add FIN and FIC tokens

SlightlyCircuitous:
Specific changes:
- Adds 13 new token entries and art for 13 reprinted tokens, as well as their reverse-relations
- Updates the oracle text of Clue and Food tokens to be in line with new conventions for tokens referring to themselves
- Corrects spelling error in Foretell reminder text
- Fixes indent on Zombie Druid token

I am really not sure why so many lines in CRANK! are marked as changed; I think they have remained untouched.

PR #308, opened 2025-06-08
Remove non-printed WOC token

tooomm:
See #303 and https://github.com/Cockatrice/Magic-Token/pull/305#issuecomment-2953562866

@SlightlyCircuitous I removed the entry completely. Do we need to add them to #85?

### DIFF
--- a/tokens.xml
+++ b/tokens.xml
@@ -16728,24 +16728,7 @@ Enchanted creature gets +1/+1 for each enchantment you control.</text>
             <token>1</token>
             <tablerow>1</tablerow>
         </card>
-        <card>
-            <name>Sorceror // Virtuous</name>
-            <text>Enchant creature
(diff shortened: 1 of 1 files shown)


SlightlyCircuitous:
No, we don't need the entry at all, really. 

The role tokens are double-faced and each role was printed on a few tokens, so Sorcerer // Virtuous is redundant with [Monster // Virtuous](https://scryfall.com/card/twoc/1/monster-virtuous) and [Monster // Sorcerer](https://scryfall.com/card/twoe/15/monster-sorcerer); all the cards that make those roles are reverse-related to those tokens already, just as a happy coincidence in how I assigned them. 

The main use of Sorcerer // Virtuous seems to be giving you a role token that is exactly correlated with the specific cards in the constructed deck they were selling. As we can use any relevant role card in Cockatrice, it's not useful to us, in addition to not existing.

PR #310, opened 2025-06-09
Rework automated version bumps

### LINKED ISSUE #220: rethink merge workflow
in https://github.com/Cockatrice/Magic-Token/pull/217 I made a mistake while using github to merge the master branch into a pr before merging that pr, because of this the xml has been broken for 15 minutes until I noticed while merging the next pr. the updating of the version number goes wrong on numerous occasions as well.

the current workflow for updating the version number works well but does not have the permissions to push to forks (this would require us to make a bot account with such permissions, it's not possible to give the default token such permissions).

it's also still possible to merge before the workflow completes, skipping the version increase. a workaround would be to use github's "auto-merge" feature which will merge after workflows complete.

instead I think the best solution is to disallow prs to be merged altogether and instead have only the workflow be able to merge prs while updating the version automatically. this works around needing a bot account as well as early merges, it does mean we'll need to keep an alternative trigger condition for this merging (eg an approval) and we could suffer some issues if the workflow implementation isn't foolproof.

tooomm:
Fixes #220

The script and comparison logic is kept unchanged, but the workflow around changed.

Pushes a new version to master branch for each change to the tokens.xml file on the same.
The workflow only runs for pushes that are not created by itself by checking the commit message to prevent loops.

### DIFF
--- a/.github/workflows/updateversion.yml
+++ b/.github/workflows/updateversion.yml
@@ -1,159 +1,83 @@
-name: Update version on pr approval
+name: Update version on tokens.xml changes
 
 on:
-  pull_request_review:
-    types: [submitted]
-  issue_comment:
-    types: [created]
-  pull_request:
-    paths: '.github/workflows/updateversion.yml'
+  push:
+    branches: [ master ]
+    paths:
+      - 'tokens.xml'
+
+concurrency:
+  group: version-bump-${{ github.ref }}
+  cancel-in-progress: false
 
 jobs:
   update_version:
-    if: >
-      ( github.event_name == 'pull_request_review' &&
-        github.event.review.state == 'approved' &&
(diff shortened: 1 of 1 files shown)


tooomm:
@ebbit1q FYI

This works like a charm when merging PR's, or pushing to master directly.
Within a few secs after the commit, a version bump happens automatically.
No more manual version bumps needed!

ebbit1q:
I considered this option in the past but found it to be suboptimal as it means all of these extra commits get made instead of squashing them into one, however my idea of having the bot push to the pr branch on review didn't always work because of permissions (the bot only is allowed access to the repo it's on) my alternative would be to make the workflow update and merge but it is more complicated and I never got around to it.

PR #314, opened 2025-07-22
Update Schema URI to https

SlightlyCircuitous:
The picture health checker threw an error about using http in https://github.com/Cockatrice/Magic-Token/pull/313 and exited abruptly. I am updating the URI to avoid the error, please let me know if there is some reason we cannot use https here. I vaguely remember there being a reason, maybe?

### DIFF
--- a/tokens.xml
+++ b/tokens.xml
@@ -4,7 +4,7 @@
     We have made a new user-editable file for token modifications.
     Currently this file is "TK.xml" and can be found in your customsets directory.
 -->
-<cockatrice_carddatabase version="4" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="https://raw.githubusercontent.com/Cockatrice/Cockatrice/master/doc/carddatabase_v4/cards.xsd">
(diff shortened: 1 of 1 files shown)


ianw11:
is this ready to get merged?
i'd love to get #313 merged so tokens work

SlightlyCircuitous:
It should be ready, yes.

ebbit1q (approved):
thank you

tooomm:
I'm pretty sure it was like that on purpose. For that reason I added a single exclude for that URL.
Maybe normal http is required to make validation pass, but I do not recall.

The link checker failing for that link was expected, see https://github.com/Cockatrice/Magic-Token/commit/c9e635da501f1f25c4bcf81f3207cc3d991aa39b.
This default w3.org schema link is going to be excluded on tool level and there is no need to explicitly list and exclude it manually anymore.
Hence I updated the arguments already, unfortunately though the release of the new lychee version with this addition is delayed as they keep including more changes.

<br>

>is this ready to get merged?
>i'd love to get https://github.com/Cockatrice/Magic-Token/pull/313 merged so tokens work

The change in this PR and the other change are independent and have no relation to each other.
Link checking is done as a good practice and quality measure.
If you look at the [failing run in the mentioned PR](https://github.com/Cockatrice/Magic-Token/actions/runs/16454680561#summary-46508602151), you clearly see that only the w3.org one is failing and all token art links are fine:
image

tooomm:
We can keep it until the upstream change in the tool and revert then.
I don't think a lot of people run validations on the file (Cockatrice does not), so not a blocker.

SlightlyCircuitous:
ah, yeah, really just hurt myself there then. I love to validate the xml.

PR #319, opened 2025-09-10
Fix Bounty Tokens

SlightlyCircuitous:
Some of the Bounty names were wrong. Not sure how that happened, but I've fixed them now.


ebbit1q:
it's very likely that these were placeholder names on scryfall while the cards were spoiled their names not fully clear, scryfall does that

PR #321, opened 2025-09-10
Fix set code for Ragavan token

SlightlyCircuitous:
Both MTGJSON and Scryfall have the _Ragavan_ token listed in DMC and the insect token listen in DMU. Updating this file to match.

edit: decided to maintain the set code for the insect token after all. see discussion below.


ebbit1q:
you're correct about the ragavan token, the insect token clearly has a dmc set symbol and not dmu, I don't know why but scryfall is just wrong here

ebbit1q:
I understand now, the issue is that there are multiple tokens that have the set DMC and the set number 7 (the other is a bear token) but have different values for the total (26 vs 12)

we're not constrained by "archival conflicts" however so I'll leave the choice up to you

ebbit1q (approved):

ebbit1q:
do we want to address that?

PR #328, opened 2025-09-18
Add missing curly braces to mana and tap symbols

SlightlyCircuitous:


ebbit1q (approved):
are these the only ones affected by this?

PR #331, opened 2025-09-19
Add missing words to oracle text

SlightlyCircuitous:
These are not really related other than they are all missing one or two words but doing three separate PRs felt like overkill.


ebbit1q (approved):
seems good

PR #332, opened 2025-09-21
Update oracle text with pronoun and target errata

SlightlyCircuitous:
Updates rules text as follows with errata introduced in the _Dominaria_ set:
- 'his or her' -> 'their'
- 'target creature or player' -> any target


ebbit1q (approved):
cool

PR #333, opened 2025-09-25
Add reminder token for warped cards

SlightlyCircuitous:
Warped cards can be cast from exile, but WotC did not print a reminder card like Adventure or Foretell cards. Users requested a way to denoted warped cards in the Discord and I have attempted to oblige.

### DIFF
--- a/tokens.xml
+++ b/tokens.xml
@@ -17863,8 +17863,8 @@ If a player casts at least two spells during their own turn, it becomes day next
             <reverse-related attach="attach" exclude="exclude">Rayblade Trooper</reverse-related>
             <reverse-related attach="attach">Red Tiger Mechan</reverse-related>
             <reverse-related attach="attach">Sinister Cryologist</reverse-related>
-            <reverse-related attach="attach" exclude="exclude">Sliver Weftwinder</reverse-related>
             <reverse-related exclude="exclude">Sliver Weftwinder</reverse-related>
+            <reverse-related attach="attach" exclude="exclude">Sliver Weftwinder</reverse-related>
             <reverse-related attach="attach">Spirited Simulacrum</reverse-related>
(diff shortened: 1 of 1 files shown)


tokens.xml
```diff
@@ -17828,6 +17828,57 @@ If a player casts at least two spells during their own turn, it becomes day next
...
             <tablerow>1</tablerow>
         </card>
+        <card>
+            <name>In the Weft</name>
+            <text>After a permanent becomes warped, you can place the exiled card here. You may cast that spell from exile.</text>
+            <prop>
+                <type>State</type>
+                <maintype>State</maintype>
+            </prop>
+            <set picURL="https://github.com/SlightlyCircuitous/image-storage/blob/main/In%20the%20Weft.png?raw=true">EOE</set>
```
ebbit1q, line 17838: this is just a redirect to https://raw.githubusercontent.com/SlightlyCircuitous/image-storage/refs/heads/main/In%20the%20Weft.png which one do we use in other tokens?

tokens.xml
```diff
@@ -17828,6 +17828,57 @@ If a player casts at least two spells during their own turn, it becomes day next
...
+            <reverse-related attach="attach">All-Fates Stalker</reverse-related>
```
ebbit1q, line 17839: we tend to be lazy and just put exclude="exclude" on every entry in this type of reminder token (see above entry), did you make sure for each of these cards whether or not they create other tokens?

ebbit1q: cool, thanks for all your effort!

ebbit1q (approved):
awesome, thank you for this addition!

tokens.xml
SlightlyCircuitous: should I change the other links as well?

PR #343, opened 2025-12-29
Add reminder card for Airbend

SlightlyCircuitous:
Cards exiled with Airbend can be cast from exile, but WotC did not print a reminder card like Adventure or Foretell cards. Users requested a way to denote such cards in the Discord and I have attempted to oblige. 

The art used is from `Aang, Airbending Master` and I have attributed the original artist; it is definitely not my art.


ebbit1q (approved):
seems good, thanks

PR #350, opened 2026-01-28
Add set num to tokens.xml

SlightlyCircuitous:
This PR adds a set number attribute (`num`) to every token printing that was physically printed in a set. Custom tokens, blog tokens, and tokens only found on Arena don't have set numbers (at least not that I can tell) and so have not received this attribute. I assigned almost all of these nums programmatically, by looking up the UUID on Scryfall. However, A Mysterious Creature, Cyberman, Morph, and Manifest are `type:Creature` and are additionally difficult to search for on Scryfall, so I have done those 15 printings manually.

There are two reasons to do this: 
- The client currently lists the set number for cards and tokens in the `Card Info` box; set number is currently always blank for tokens and this would fill those fields in
- Lookup by set num was requested in https://github.com/Cockatrice/Magic-Token/pull/340#issuecomment-3618802481 and this provides preliminary XML support in the event this is implemented in the client

I am currently uncertain what (if anything) to do about tokens that do not have an official set number and open to suggestions, but this PR at least solves the above issues for the majority of the file.

### DIFF
--- a/tokens.xml
+++ b/tokens.xml
@@ -1943,11 +1943,6 @@ Whenever this creature attacks, target attacking creature gains flying until end
                 <cmc>0</cmc>
                 <pt>1/1</pt>
             </prop>
-            <set picURL="https://cards.scryfall.io/large/front/9/0/902fba98-92b7-461b-ac6e-29cb2c4e307f.jpg?1738355211" uuid="902fba98-92b7-461b-ac6e-29cb2c4e307f">DFT</set>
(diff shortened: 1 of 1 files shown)


SlightlyCircuitous:
Rebased to master to resolve conflicts.

SlightlyCircuitous:
For whatever reason, this added a couple duplicate lines. reviewing the PR locally to try and find them all.

PR #356, opened 2026-03-10
Add Wall token to Rammas Echor

SlightlyCircuitous:
Reported in the Discord.

This token was not reprinted in LTR/LTC so its relation was overlooked in https://github.com/Cockatrice/Magic-Token/pull/227.


ebbit1q (approved):
thanks!

PR #360, opened 2026-04-14
Add SOS and SOC Tokens

SlightlyCircuitous:
This PR makes the following changes:

* Adds token entries and reverse-relations for 7 new tokens
* Adds set lines and reverse-relations for 31 unique reprints
* Remove blank line
* Add missing `maintype` to Shredder, Foot Clan Overlord (missed in https://github.com/Cockatrice/Magic-Token/pull/354)


ebbit1q (approved):
thank you!

PR #365, opened 2026-05-28
Add all prinitings of TLA tokens

SlightlyCircuitous:
Since the introduction of the printing selector for tokens, it is now becoming standard to include all printings for each token in a set. TLA predates this change so I am adding them now. This was additionally requested on our Discord.


ebbit1q (approved):
ty

PR #366, opened 2026-06-12
Add all printings for EOE tokens

SlightlyCircuitous:
Since the introduction of the printing selector for tokens, it is now becoming standard to include all printings for each token in a set. EOE predates this change so I am adding them now. More cool spaceships for everyone.


ebbit1q (approved):
thanks!

PR #382, opened 2026-08-20
Add MBC tokens

SlightlyCircuitous:
* Adds reverse-relations for extant tokens that are made by new cards in MBC
* Adds 3 new token entries for new tokens made by new cards in MBC that don't have printings (yet?)
* Does not touch cards that make copies of existing cards (e.g. Nivea), pending discussion on how best to implement these cards

(Note that this set isn't out until November, but it seems like all the new cards are on Scryfall already and this was requested in the Discord so I am doing them early. With any luck, we will get some tokens with this set and I can update the art in a later PR.)


ebbit1q (approved):
thanks, when will we see mbc token scans? at release date? nov 13

SlightlyCircuitous:
> thanks, when will we see mbc token scans? at release date? nov 13

If they are getting printed, we should see them on Scryfall about 2 weeks before release. So Oct 30 or so.

PR #380, opened 2026-08-15
Add Suspend state token

SlightlyCircuitous:
Requested in the Discord

* Adds art for a Suspend reminder token (based on [this public domain painting](https://commons.wikimedia.org/wiki/File:Else_M%C3%BCller-Kaempff_Stillleben_mit_Sanduhr,_Kannen_und_Essgeschirr.jpg))
* Adds token entry and reverse-relations for all cards with Suspend

Please let me know if I missed any cards or added too many. There are a lot of cards that interact with Suspend!

PR #388, opened 2026-08-31
Differentiate suspend state from suspend card

SlightlyCircuitous:
Fixes an issue caused by https://github.com/Cockatrice/Magic-Token/pull/380
`Suspend` is already a card name, in addition to a keyword. 

Changes the name of the state token from `Suspend` to `Suspended`.


ebbit1q:
thank you for noticing this

PR #389, opened 2026-08-31
Suspend entry fixes

SlightlyCircuitous:
Builds on https://github.com/Cockatrice/Magic-Token/pull/380

Changes

- Removes `Amy Pond` as it does not Suspend. 
- Adds planes `Amy's Home` and `Talon Gates` that can suspend other cards


ebbit1q (approved):
thank you!

